PR #17, opened 2013-01-07
World of warcraft required empty libs

tizbac:


LubosD:
Not relevant any more.

PR #60, opened 2014-02-12
LIBDIR-FIX: replace hardcoded lib{SUFFIX} with {CMAKE_INSTALL_LIBDIR}

thopiekar:
Well, as I found out by packaging darling. The main CMakeFile looks for lib32/lib64 in CMAKE_INSTALL_LIBDIR and sets SUFFIX, but the CMakeFiles in the subdirectories are using lib{SUFFIX} again. Thats why CMAKE_INSTALL_LIBDIR is in some way "overwritten".

My commit is untested, but I think it will fix it.


ahyattdev:
I think that the purpose of this PR has already been fulfilled.

PR #126, opened 2016-02-26
Improvements for darling script

X0rg:
Hi.

There are some changes I have made to avoid errors with dyld (like checking for /dev/mach permissions).
I have added functions to run commands as root (like modprobe, ...), inspired by [yaourt](https://github.com/archlinuxfr/yaourt).


LubosD:
Thanks a lot!

Issue #134, opened 2016-03-04
installing kernel module fails


Elronnd:
(Ubuntu 14.04 LTS)

Darling itself build and installed fine, and the kernel module compiles fine, but when I try to install it, I get:

    $ sudo make install
    make -C /lib/modules/3.19.0-51-generic/build M=/home/elronnd/darling/src/lkm modules_install
    make[1]: Entering directory `/usr/src/linux-headers-3.19.0-51-generic'
      INSTALL /home/elronnd/darling/src/lkm/darling-mach.ko
    Can't read private key
      DEPMOD  3.19.0-51-generic
    make[1]: Leaving directory `/usr/src/linux-headers-3.19.0-51-generic'

LubosD:
This is just a warning. You should be able to load the module just fine.

Elronnd:
But if I then do `modprobe darling-mach`, it complains that it can't find it.  However, the module still seems to have been installed in `/lib/modules/3.19.0-51-generic/extra/darling-mach.ko`.  Maybe I need to let modprobe know it has to look there?

LubosD:
This is normally done by `depmod`. But you should always be able to `insmod` the module with an absolute path.

Elronnd:
That worked.  But now, it's looking for stuff in `/libexec` which doesn't exist on ubuntu -- but `/usr/local/libexec` contains a `darling/` dir, but just that, so I moved `/usr/local/libexec` into /, but there's a host of other problems.  It complains about a missing `/bin/darling`, lack of `realpath`, and then segfaults.

LubosD:
I'm sorry you're having this experience. Unfortunately, every Linux distribution seems to have a ton of quirks. I regularly test Darling on Debian and I hoped things would be similar enough on Ubuntu.

Until I figure this out, I suggest re-running cmake with `-DCMAKE_INSTALL_PREFIX=/usr/local`, followed by make and make install.

Elronnd:
It's giving me a weird error now

    $ darling shell
    /usr/local/bin/darling: 36: [: Illegal number: 
    terminate called after throwing an instance of 'std::regex_error'
      what():  regex_error
    zsh: abort (core dumped)  darling shell

And it looks like the line that's being complained about is

            if [ "$updatets" != "disable" -a "$updatets" -lt $selfmtime ]; then

LubosD:
I will probably force Bash for this script.

[...]

PR #135, opened 2016-03-05
Fix timestamp check in darling script (see #134)

X0rg:
When I put "disable" or "disabled" in $DPREFIX/.update-timestamp, I've got one error.
Sh doesn't like to compare a string with an integer.

### DIFF
--- a/src/dyld/darling
+++ b/src/dyld/darling
@@ -33,8 +33,8 @@ check_prefix() {
     else
 
         updatets=$(cat $DPREFIX/.update-timestamp)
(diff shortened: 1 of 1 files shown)


LubosD:
OK, but I'll extend the code to be even more robust.

X0rg:
Ok. Maybe we need to ensure that $updatets is an integer before doing comparison with $selfmtime.

X0rg:
Done in dcb78a7, thanks. :)

PR #137, opened 2016-03-18
Changes needed to build on Debian system

jcoffland:
Note ``bootfiles.h`` is an OSX file not current found in darling or any of its submodules.


jcoffland:
thanks

PR #177, opened 2016-11-06
Initial ruby submodule

ahyattdev:
I also sent a transfer request to @LubosD for the darling-ruby repo currently on my account. If this PR is accepted, you can then accept the transfer request and then transfer it to darlinghq. GitHub doesn't allow requesting a transfer to an organization unless you are an admin for some reason. 

Also, a commit that updates the corecrypto submodule sneaked into this PR.


LubosD:
Thanks a bunch!

Issue #157, opened 2016-09-23
Build issue on Arch Linux x86_64


LeRieur:
It doesn't seem to be related to the packaging: https://aur.archlinux.org/packages/darling-git https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=darling-git

Here's the output of makepkg:
https://0bin.net/paste/2Xv8WyeK8VkIcwoN#228SBG-bwAb+1F93Qm4qN43c1CZhL5P8ZvB59kkK4Qj

LubosD:
You lack /usr/bin/arch, which could cause some problems. The final crash, however, is yet another compiler incompatibility. I'll take a look at it.

LubosD:
Umm, checking it again, it looks more like a bug either in your version of LLVM or Binutils. Downgrading binutils could help. See https://groups.google.com/forum/#!topic/llvm-dev/vgVAb2B9DVg

LeRieur:
Downgrading binutils to 2.26.1 fixed it. Thanks.

/usr/bin/arch is not part of util-linux since 2.23. Even without it, it seems to build fine.

Maxqia:
:P I tried to get /usr/bin/arch included in Arch Linux, didn't go so well

tpimh:
I finally managed to compile it on Arch Linux.

I used the following workarounds:

1. Creating executable `/usr/bin/arch` with the following content:

    #!/bin/sh
    echo x86_64
2. Downgrading binutils as mentioned above
3. `sed -i 's/-Werror//' src/external/libkqueue/CMakeLists.txt` (build breaks because of expansion-to-defined warning)
4. `cd src/libunwind; curl -qsL http://github.com/JuliaLang/libosxunwind/pull/7.patch | patch -p1` (build breaks because of use of COMPILE_TIME_ASSERT)

LubosD:
I'll see what I can do about the `/usr/bin/arch` usage, but I'm surprised Arch Linux doesn't have it - Debian (Testing) does, ~amd64 Gentoo too...

Scimmia22:
/usr/bin/arch is not enabled by default by coreutils upstream. Assuming it's present on all Linux systems is a bad idea.

Maxqia:
Reopened issue here : https://bugs.archlinux.org/task/49890 ...

tpimh:
Maybe `uname -p` is an option?

LubosD:
[...]

PR #181, opened 2016-11-08
add build dependencies to README

GlenDC:
Could the added arch linux dependencies perhaps also be added to the website? I couldn't seem to find the repository for the website, otherwise I would have helped out with that.

For now I can't build the i386 build on Arch Linux (x64) though, as I get stuck at the following during `make`:

```
$ make
[  0%] Built target migcom
[  0%] Built target system_duct
[  4%] Built target emulation
[  4%] Building C object src/kernel/libsyscall/CMakeFiles/libsyscall.dir/custom/errno.o
/tmp/errno-4a1da7.s: Assembler messages:
/tmp/errno-4a1da7.s:630: Error: inconsistent uses of .cfi_sections
clang-3.9: error: assembler command failed with exit code 1 (use -v to see invocation)
make[2]: *** [src/kernel/libsyscall/CMakeFiles/libsyscall.dir/build.make:1611: src/kernel/libsyscall/CMakeFiles/libsyscall.dir/custom/errno.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:515: src/kernel/libsyscall/CMakeFiles/libsyscall.dir/all] Error 2
make: *** [Makefile:150: all] Error 2
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,26 @@ For complete instructions, visit [DarlingHQ.org](http://www.darlinghq.org/build-
 <img src="http://teamcity.dolezel.info/app/rest/builds/buildType:(id:Darling_DebianStableX8664)/statusIcon" title="Debian stable build for x86-64"/>
 </a>
 
+Required dependencies
+
+Debian (stable):
+
+```
+$ sudo apt-get install cmake clang bison flex xz-utils libfuse-dev libxml2-dev libicu-dev libssl-dev libbz2-dev zlib1g-dev libudev-dev linux-headers-amd64
+```
+
+Ubuntu (16.04):
+
+```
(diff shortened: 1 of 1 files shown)


LubosD:
Take a look here, I think it's a bug in binutils: https://github.com/darlinghq/darling/issues/157#issuecomment-249292978

I'll publish the website's repo.

LubosD:
https://github.com/darlinghq/www.darlinghq.org

GlenDC:
Ah ok that seems to talk about a x86_64 build, which I can build successfully on Arch Linux. It's only when trying to build the i386 build that it fails. I'll check out the issue out in more detail later though.

PR #182, opened 2016-11-11
Add Python to /etc/darling/dylib.conf, move the Ruby folder up, and add DebugSymbols.framework

ahyattdev:
Question:

How does Darling handle it when files such as /etc/darling/dylib.conf are changed in the source of the project which also exist in the prefix? I couldn't get it to install and had to copy it over to my prefix.


LubosD:
Unfortunately, this part of prefixes (stuff such as dylib.conf) hasn't yet been fully thought through :-( You are absolutely welcome to make suggestions how it should be improved.

PR #185, opened 2016-11-12
Add sandboxed home directory

kamillys:
Signed-off-by: kamil <kamillys@gmail.com>

### DIFF
--- a/src/dyld/darling
+++ b/src/dyld/darling
@@ -66,8 +66,19 @@ setup_prefix() {
     rm -f "$1/dev"
     ln -sf "system-root/dev" "$1/dev"
 
+    rm -rf "$1/sandboxed-home"
+
     rm -f "$1/home"
-    ln -sf "system-root/home" "$1/home"
+    if [ "${DARLING_SANDBOXED_HOME}" != "1" ]; then
+        >&2 echo "Setting sandboxed home"
+        mkdir "$1/sandboxed-home"
+        ln -sf "sandboxed-home" "$1/home"
+        for USERDIR in $(ls "$1/system-root/home/"); do
+            mkdir "$1/sandboxed-home/$USERDIR"
+        done
+    else
+        ln -sf "system-root/home" "$1/home"
+    fi
 
     rm -f "$1/Users"
     ln -sf "home" "$1/Users"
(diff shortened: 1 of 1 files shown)


src/dyld/darling
```diff
@@ -66,8 +66,19 @@ setup_prefix() {
...
 
+    rm -rf "$1/sandboxed-home"
+
     rm -f "$1/home"
-    ln -sf "system-root/home" "$1/home"
+    if [ "${DARLING_SANDBOXED_HOME}" != "1" ]; then
+        >&2 echo "Setting sandboxed home"
+        mkdir "$1/sandboxed-home"
+        ln -sf "sandboxed-home" "$1/home"
+        for USERDIR in $(ls "$1/system-root/home/"); do
```
stek29: Shouldn't this be `"/system-root/home/"` and not `"$1/system-root/home/"`?
Also, this will cause troubles if there are directories containing space in name.
So I'm sure this should be:
  `for USERDIR in /system-root/home/*; do`

src/dyld/darling
```diff
@@ -66,8 +66,19 @@ setup_prefix() {
     rm -f "$1/dev"
     ln -sf "system-root/dev" "$1/dev"
 
+    rm -rf "$1/sandboxed-home"
+
     rm -f "$1/home"
```
msva: I'd say this function (`setup_prefix`) **REQUIRES** a rewrite with lot more checks. For example, for emptyness of `$1` (nobody should **never** do anything like `rm -f ${variable}/<path_on_filesystem>`. It can SUDDENLY appear the case of empty variable in the future, which lead to bad event) and if it is really necessarily to rm'ing that files (or they all already fine).

So, if you really want rm'ing the path prepended by variable, check for variablr

LubosD:
It'd great if you could fix what the reviewers found. At any rate, the final plan (so that you know) is to automatically bind mount various directories into the chroot based on `/etc/fstab` inside the chroot.

src/dyld/darling
kamillys: True. However if any mkdir fails, function returns prematurely.
I'll add checks for $1 and replace $1 usage with local variable.

PR #189, opened 2016-11-15
Fix and hide some warnings

ahyattdev:


LubosD:
Hi, thanks for the PR and sorry for the mess. I created a better fix for the problem (that your 2nd commit tries to resolve) and manually merged just the 1st commit.

PR #191, opened 2016-11-16
Create a submodule for man

ahyattdev:
It is already at the point where it builds and links the `man` executable. Some work is needed to get `/etc/man.conf` in place and build some other necessary utilities. 

@LubosD I sent a transfer request for the `darling-man` repository on my account to your's, to transfer it to @darlinghq.


LubosD:
Thanks a lot!

PR #192, opened 2016-11-17
Update man and remove dtrace dependency

ahyattdev:
# man
This update has a `man` that just needs some more dependencies to work, including:

`/usr/bin/groff`
`/usr/bin/tbl`

It also installs some manpages for `man` itself

# ruby
Make sure that you clean the build folder for `src/external/ruby`, because it will pick up the old, generated `probes.h` (it searches `CMAKE_BINARY_DIR` before `CMAKE_SOURCE_DIR`) instead of the new one. 

A Hello World program works now!

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -94,7 +94,7 @@ add_subdirectory(CommonCrypto)
 add_subdirectory(csu)
 add_subdirectory(external/python/2.6/Python-2.6.9)
 add_subdirectory(external/python/2.7/Python-2.7.10)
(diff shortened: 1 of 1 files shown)


LubosD:
From now on, ruby is needed to build Darling. What do you think about also bundling pregenerated .tmpl files?

ahyattdev:
I think that would be smart, since portability doesn't matter this point, except between x86_64 and i386 Linux. 

Using Ruby scripts to generate files necessary to build Ruby wasn't the best move of their's. 😆

PR #198, opened 2016-11-22
Build the manpage database when the prefix is set up

ahyattdev:
A step to `setup_prefix` in `src/dyld/darling` that runs the script to build the manpage database. 

The manpage database (stored in `/usr/share/man/whatis`) is used by commands that search manpages, such as `apropos` and `whatis`.

### DIFF
--- a/src/dyld/darling
+++ b/src/dyld/darling
@@ -99,6 +99,9 @@ setup_prefix() {
 
     install_pkg "org.darlinghq.pkg.OpenSSLCertificates"
 
+    >&2 echo "Building manpage database..."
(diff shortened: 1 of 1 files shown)


ahyattdev:
There is a regression where the script exists after updating the manpage database. This is not due to an error because it successfully builds it. Running `darling shell` a second time brings up `bash` no problem.

ahyattdev:
It turns out that this is because `exec` replaces the shell with that command. Is there any way to execute `/usr/libexec/makewhatis` so `makewhatis` operates with the prefix as the root file system?

LubosD:
As long as DPREFIX is exported, this should be the case. You just need to specify full (real) path to makewhatis.

ahyattdev:
The solution was running the command without `exec`. Duh!

Issue #197, opened 2016-11-21
New Darling prefix design


LubosD:
* Replace `darling` script with a C binary
  * Detects if the prefix is already "booted" and if not, initiates boot.
* Make OpenSSL certificates part of Darling (instead of installed package).
* Build prefixes on overlayfs rather than on copying files.
* Implement `prelaunchd` as the init process for prefixes. `prelaunchd` creates a mount and pid namespace (consider setting up a UID namespace - do we need that?).
  * When we have working launchd, prelaunchd will exec() launchd after its work is done. Until that is done, prelaunchd will end in a loop reaping child processes.
* Abandon `dylib.conf`.
  * Use symlinks? Or generate dummy Mach-O files that just link e.g. `@darling/libFoundation.so`?

ahyattdev:
I agree that the `darling` shell script needs to be replaced. It is taking on more than what was originally intended. 

Not having to copy files in `/usr/local/libexec/darling` to `~/.darling` because of an overlay would be great. It would solve changes to the `libexec` files not being synced to the prefix. 

Would using symlinks instead of `dylib.conf` work since the prefix is `chroot`ed?

Broad question, but why not generate real Mach-Os?

LubosD:
> Would using symlinks instead of dylib.conf work since the prefix is chrooted?

It would work, because VirtualPrefix.cpp in libdyld has special handling for symlinks. But generating semi-real Mach-O files (referring internally to an ELF file with actual implementation) somehow feels better.

> Broad question, but why not generate real Mach-Os?

Using Mach-O for Darling's libraries and executables would eliminate many hacks, but there is a catch. We sometimes need to link our frameworks against system Linux libraries. Currently, this is the case only for AudioUnit I think, but in future this will also apply to CoreGraphics, AppKit etc.

In these cases, [Apple's ld](https://opensource.apple.com/source/ld64/) (which we would need to build) would need to somehow link to ELF files. Either we could add a simple ELF parser into ld64 or feed it dummy Mach-Os containing the symbols from the real ELF file.

It is probably viable (and likely even preferable in long term), but for now I took the easy way out by not building Mach-Os.

ahyattdev:
Having dummy Mach-Os but real symbols in them would allow us to use Linux libraries and also allow a compiler running in the prefix to link to the Mach-Os. This solution allows using Darling for development without the macOS SDK.

X0rg:
[...]

PR #205, opened 2016-11-27
Add a submodule for OpenSSL Certificates

X0rg:
Related to #197.
It adds a submodule with OpenSSL Certificates.

OpenSSL Certificates are installed in `${CMAKE_INSTALL_PREFIX}/libexec/darling/System/Library/OpenSSL/certs`.
Symbolic links are created by CMakeLists.txt.

PS: I'm sorry, I'm not sure of way to add submodules. Tell me if I'm wrong.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -180,3 +180,6 @@
 [submodule "src/external/bc"]
 	path = src/external/bc
 	url = ../darling-bc.git
+[submodule "src/external/openssl_certificates"]
+	path = src/external/openssl_certificates
+	url = ../darling-openssl_certificates.git
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -106,6 +106,7 @@ add_subdirectory(sandbox)
 add_subdirectory(Cocoa)
 add_subdirectory(external/file/file)
 add_subdirectory(external/libxpc)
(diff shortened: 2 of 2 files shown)


LubosD:
The submodule stuff is OK.

One question: I see you use openssl to generate certificate names. That's great, BUT the hashing algorithm has changed between 0.9.x and 1.something (from md5 to sha1 IIRC). MacOS is stuck with 0.9.x.

Could you please check that your cmake script produces the same hashes as what is inside of http://packages.darlinghq.org/org.darlinghq.pkg.OpenSSLCertificates/org.darlinghq.pkg.OpenSSLCertificates-1.pkg ?

If not, try doing something like `openssl x509 -noout -modulus -in certificate.crt | openssl md5` instead, maybe it will generate the right hash.

X0rg:
Indeed, hashes are differents, example with ACCVRAIZ1.crt:
3c9a4d3b (package) / a94d09e5 (openssl hash)
But when I use `openssl x509 -noout -subject_hash_old -in`, hash is correct.

LubosD:
Even simpler then, thanks!

X0rg:
Yes, using `openssl md5` doesn't produce the right hash in fact.

PR #207, opened 2016-11-27
Support module loading from new darling executable

X0rg:
Example:
```
$ getcap /usr/bin/darlingC 
/usr/bin/darlingC = cap_sys_module+ep

$ lsmod | grep mach       
     
$ /usr/bin/darlingC shell
Cannot unshare(CLONE_NEWNS): Operation not permitted

$ lsmod | grep mach
darling_mach           77824  0
```


X0rg:
About `unloadKernelModule()`, I don't know if we need to unload module when darling exit with non-zero status.
If yes, we should probably add a function/macro to cleanup before exiting, avoiding adding  `unloadKernelModule()` before all `return 1` and `exit(1)`.

LubosD:
I'm against module unloading, there can be many darling processes running. Automatic loading is only a convenience feature.

X0rg:
Ok. Did I need to revert 9e58414? Or can we assume this executable will take options, like `darling unload`?

LubosD:
I'll probably take options, so I'll merge it and just uncomment default unloading.

X0rg:
Ok, thanks. :)

PR #210, opened 2016-12-04
Ignore -Wexpansion-to-defined warning

X0rg:
Close #204

PR #211, opened 2016-12-04
Fix unknown warning option '-Wno-expansion-to-defined'

X0rg:
Related to #210


ahyattdev:
I tweaked the conditional so it is `NOT LESS_THAN` instead and manually merged this.

PR #217, opened 2016-12-30
Implement the new 'darling' executable

bugaevc:
Part of #197:

* Replace darling script with a C binary
    * Needs SETUID to load the kernel module and for namespace/mounts manipulation.
    * Creates and populates the prefix if it doesn't exist.
    * Detects if the prefix is already "booted" and if not, initiates boot.
    * If there's no prelaunchd/darling-init running, launches it first.
    * Spawns the child process (e.g. the shell) into the PID/UID/mount namespaces of prelaunchd.
    * Sets up some environment for the child (i.e. DPREFIX, PATH, HOME, working directory) - ideally, we'd need something like systemd's `machinectl shell` to spawn the new process from PID 1 and not `darling` executable, yet make it transparent for the user.
    * Waits for the child to exit and returns its status.
* Build prefixes on overlayfs rather than on copying files.
* Implement prelaunchd (`darling-init`) as the init process for prefixes. prelaunchd creates a mount, user and PID namespaces. Until we have a working launchd, prelaunchd ends up looping and reaping child processes.

To build the new executable, pass `-DNEW_DARLING=TRUE` to `cmake`.

Most of the code runs as root (SETUID), so we must be extra careful about the security:
* Check FS permissions.
* Check that the process whose PID is saved in `$DPREFIX/.init.pid` actually belongs to the current user -- otherwise it'd be possible to attach to someone else's namespace.
* Carefully check buffer sizes and only use *n* versions of `scanf`&`printf` to prevent buffer overrun.

Also, I fixed a couple bugs in the previous version of the code, e.g. taking or not taking the trailing newline (`\n`) into account when reading text from files.

Some stuff (for example, `hdiutil`) predictably doesn't work. I'm not sure that mapping UIDs/GIDs as "0 uid 1" is what we need.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,15 @@ make
 make install
 ````
 
+To build the new, experimental `darling` executable:
+
+````
+cmake ../.. -DCMAKE_TOOLCHAIN_FILE=../../Toolchain-x86_64.cmake -DNEW_DARLING=TRUE
+make
+make install
+chmod +s $(which darling)
+````
+
 #### For running i386 OS X binaries
 
 <a href="http://teamcity.dolezel.info/viewType.html?buildTypeId=Darling_DebianStableX8664&guest=1">
--- a/README.md
+++ b/README.md
@@ -164,7 +164,7 @@ Congratulations, you have just compiled and run your own Hello world application
 
 ### AppKit
 
-AppKit is still highly expiramental and incomplete, but to work on it you need to configure CMake with `-DFRAMEWORK_APPKIT=1` and install some additional packages. 
+AppKit is still highly expiramental and incomplete, but to work on it you need to configure CMake with `-DFRAMEWORK_APPKIT=1` and install some additional packages.
 
 Ubuntu 16.04:
 ```
--- a/README.md
+++ b/README.md
@@ -164,7 +164,7 @@ Congratulations, you have just compiled and run your own Hello world application
 
 ### AppKit
 
-AppKit is still highly expiramental and incomplete, but to work on it you need to configure CMake with `-DFRAMEWORK_APPKIT=1` and install some additional packages.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -164,7 +164,7 @@ Congratulations, you have just compiled and run your own Hello world application
 
 ### AppKit
 
-AppKit is still highly expiramental and incomplete, but to work on it you need to configure CMake with `-DFRAMEWORK_APPKIT=1` and install some additional packages. 
+AppKit is still highly expiramental and incomplete, but to work on it you need to configure CMake with `-DFRAMEWORK_APPKIT=1` and install some additional packages.
```
thopiekar: Typo here. `expiramental` -> `experimental`

bugaevc:
Notes & questions:
* The only way to cause `darling-init` to quit for now is to `kill -9` it.
* Is there a reason why we don't use `binfmt_misc`?
* How do we handle root/non-root inside vs root/non-root outside the UID namespace?
* I'd like to get rid of VirtualPrefix and mount DPREFIX directly onto `/`
* Maybe make the whole `/Volumes` directory a FUSE and mount it on namespace creation? Somehow combine it with `/run/media/$USER/`?

README.md
bugaevc: Ok, will fix in a moment (although the text wasn't written by me).

bugaevc:
We're remounting `/` as a "slave" (it was "private" previously) to propagate mounts that are done on the system after PID 1 was launched (a common use-case would be removable drives).

ahyattdev:
Thanks! I would merge, but @LubosD should make that decision since it is a big change. I added him as a reviewer to the PR.

bugaevc:
Should we put all the initial prefix contents (that are now created in `setupPrefix()`) into `/usr/local/darling/libexec` as well, to make initial prefix set-up (almost) a no-op?

By the way, as you can see in the code, I got rid of `/home` and made `darling` executable automatically detect and replace `/home/path` with `/Users/path` for `HOME` and CWD.

Issue #221, opened 2017-01-04
Using Mach-O's for Darling itself


LubosD:
The following idea is currently tested in the [using-machos-experiment](https://github.com/darlinghq/darling/tree/using-machos-experiment) branch.

- Build all libraries & frameworks in Darling as Mach-O objects (.dylibs). This actually means reverting many changes made in assembly code.
- Build all (prefix) executables in Darling as Mach-O executables.
- Try using original dyld for loading binaries. This would sadly mean dropping our own libdyld.
  - Provide a very simple (ELF) executable capable of loading dyld itself (dyld would already be a Mach-O) and handing control over to it.
  - This thin loader would provide pointers to ELF loader's APIs (dlopen, dlclose, dlsym, ...) somewhere.
  - We would also need to be able to build libc & libsystem_kernel specifically as a static library for dyld. There is a special define to be used for the libc build.
- Once we need to call "foreign APIs" (e.g. ALSA, OpenGL, ...), provide a generator that creates a Mach-O wrapper for any given ELF library, creating simple stubs that automatically invoke ELF dlsym on first use and forward calls to the real implementation. Basically something similar to `dyld_stub_binder`, except this would be `dyld_elf_stub_binder`.
- We will also need to drop libobjc2 and use Apple's runtime instead. There have already been some tests of this in the apple-objc4-experiment branch.

LubosD:
Now I've built most of the stuff, except libobjc (will use Apple's objc4) and frameworks in general.

Most things work, but `/bin/bash` and `/bin/stat` crash due to an incorrect `free()` call. The biggest problem is debugging, because `gdb` is unable to load debug info for any of the Mach-Os, so it's a nightmare.

Therefore, the next step is to get gdb working.
Good news:
* Multitarget builds (`USE=multitarget` on Gentoo) of gdb include support for Darwin and Mach-O. They look for `_dyld_all_image_infos` to communicate with dyld.
  * Debian's gdb-multiarch doesn't include this support, but we can of course build gdb manually (and provide build instructions).
  * I haven't yet successfully used it, so maybe vanilla code will not work. In that case we'll just fork gdb.

Bad news:
* The last time dyld supported gdb was in OS X 10.8, it was later replaced with support for LLDB.
* LLDB is completely shite on Linux.
  * I will forward port the gdb integration.

LubosD:
[...]

Issue #220, opened 2017-01-03
New /Volumes design


bugaevc:
This is a design proposal and a request for comments.

On macOS, `/Volumes` is a directory that holds mountpoints for all the drives in the system, as well as for mounted DMG files, somewhat like `/run/media/$USER` on Linux. On Darling, this directory is currently only used for mounting DMGs.

To make Darling better resemble macOS, we should put the following under `/Volumes`:

* Static items
    * `/Volumes/DarlingEmulatedDrive` -- a symlink to `/`. This is what macOS does for the root drive.
    * `/Volumes/SystemRoot` -- this is where host's `/` gets mounted (this was previously `/system-root`, but putting it in `/Volumes` feels much more true-to-the-spirit).
* DMG mountpoints for `hdiutil`. This works great currently, but it's not so easy in the new containerized design (see #197), because we can't mount FUSE from inside user namespaces. The solution is to either abandon user namespaces (do we really need them?) or to enhance `hdiutil` to e.g. join the initial namespace before doing its work.
* Rest of the drives on the host -- basically, we should add everything from `/run/media/$USER` here. Unfortunately, we can't use overlayfs, because it doesn't support modifying the underlying files while mounted, which is what happens to host's drives all the time. We can either use smth like [mergfs](https://github.com/trapexit/mergerfs) or write a daemon that watches `/run/media/$USER` for changes and creates/removes symlinks in `$DPREFIX/Volumes`. WHichever we choose, these folders won't show up as mount points. We could bind-mount them to the original ones instead of symlinking, but this may interfere with udisks unmounting them.

ahyattdev:
Does (this comment)[http://unix.stackexchange.com/questions/239416/how-to-modify-a-lower-file-through-an-overlayfs-directory#comment410720_239416] help with modifying the underlying files?

bugaevc:
@ahyattdev No. They are talking about modifying files through the overlay (that would be from inside darling container in our case), which would work if we made `/run/media/$USER` the upper directory. What wouldn't work is modifying the files on the `/run/media/$USER` directly (that would be from the host Linux system) while it's mounted as a part of overlay. [The documentation](https://www.kernel.org/doc/Documentation/filesystems/overlayfs.txt) is rather clear here:

> Changes to the underlying filesystems while part of a mounted overlay filesystem are not allowed.  If the underlying filesystem is changed, the behavior of the overlay is undefined, though it will not result in a crash or deadlock.

[...]

PR #222, opened 2017-01-05
Get of rid of the VirtualPrefix

bugaevc:
This is my work on building the new layout of the "Darling container" (see #197).

* Deprecate the old 'darling' script and always build the new executable.
    * One notable maybe-missing thing is OpenSSL certificates -- please look into this.
* Get rid of Darling's chroot-emulation-in-syscall-emulation (known as VirtualPrefix). Instead, we mount `$DPREFIX` directly onto `/` inside the container.
    * Mount a new `procfs` (for our PID namespace) onto `/proc` inside.
    * This means that native executables (accessible through `/Volumes/SystemRoot/...`) see the new layout, too. You can use e.g. `/Volumes/SystemRoot/bin/ps`.
    * The complicated part here is loading shared libraries.
        * Set up `/etc/ld.so.conf.d` and run `ldconfig` at installation time.
        * Pass `LD_LIBRARY_PATH` for `/usr/local/lib/darling`.
            * This is because I couldn't get it to work with `/etc/ld.so.conf.d/darling.conf`; this won't be hopefully needed after the Great Mach-O Bringup (see #221).
        * Symlink `/lib -> /Volumes/SystemRoot/lib` and `lib64 -> /Volumes/SystemRoot/lib64`.
            * This is because most ELFs have `/lib64/ld-linux-x86-64.so.2` (or 32bit version) hard-coded as their interpreter.
* Create most of the prefix filesystem layout inside `/usr/local/libexec/darling` at installation time.
* On `/Volumes` (see #220):
    * Move `/system-root` to `/Volumes/SystemRoot`.
    * Turn off user namespaces for now -- they were not really useful anyway.
    * `hdiutil` works if launched like `DPREFIX=/ PATH=/Volumes/SystemRoot/bin /usr/bin/hdiutil`. It should be fairly easy to make it support the new layout.
* Misc changes that are better described by their commit messages and diffs.

Please do review and test!


LubosD:
There is one little problem (but it can be solved after I review the changes and merge the code) - in some cases, a little bit of VirtualPrefix code (or simply new code) are still needed:

HFS+ is case insensitive and apps depend on this. I've seen software explicitly checking whether the filesystem where it's installed is case sensitive (usually HFSX on macOS) and then refusing to run.

So we will need to have a special resolution algorithm, which on the other hand only needs to be triggered in case the given system call returns -LINUX_ENOENT.

LubosD:
One more question before I merge (and I may be a little confused, possibly): where do we chroot?

bugaevc:
> One more question before I merge (and I may be a little confused, possibly): where do we chroot?

We don't actually use `chroot()`, instead, we [`pivot_root`](https://linux.die.net/man/2/pivot_root) right after mounting overlayfs onto the prefix.

Issue #231, opened 2017-01-13
CMAKE_LIBRARY_ARCHITECTURE is empty


bugaevc:
I may be misunderstanding everything big time, but it seems to me that:

* On Debian-based x64 systems, Darling libs are supposed to go under `/usr/local/lib/x86_64-linux-gnu/darling`
* In `Toolchain-x86_64.cmake`, `CMAKE_INSTALL_LIBDIR` is set to `"lib/${CMAKE_LIBRARY_ARCHITECTURE}"`
* In `darling-config.h.in`, `LIB_PATH` is `#define`d to be `"${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}/darling"`, i.e. `/usr/local/lib/x86_64-linux-gnu/darling`
* We pass `SYSTEM_ROOT LIB_PATH`, i.e. `/Volumes/SystemRoot/usr/local/lib/x86_64-linux-gnu/darling` to the container in `$LD_LIBRARY_PATH`.

However, from what I see in practice, libs end up in `/usr/local/lib/darling` and `$LD_LIBRARY_PATH` gets set to `/Volumes/SystemRoot/usr/local/lib//darling` (notice the double slash; confirmed by @ahyattdev in #223).

**This works fine for x86_64, but the libs get overwritten if you try to compile for i386, because it uses the same path.** Basically, i386 build is broken (builds, but doesn't work), at least for me.

The reason for this seems to be that `CMAKE_LIBRARY_ARCHITECTURE` is sometimes not set/empty, as mentioned elsewhere on the internet (ex. TimothyGu/libilbc#10, [this one about Fedora](https://lists.freedesktop.org/archives/beignet/2015-March/005366.html)). Thus we get `/usr/local/lib//darling`.

I'm not proficient enough in CMake to suggest a solution (maybe just hardcode `x86_64-linux-gnu`/`i386-linux-gnu` instead of `CMAKE_LIBRARY_ARCHITECTURE`?).

Also, we should pass both `/Volumes/SystemRoot/usr/local/lib/x86_64-linux-gnu/darling` and `/Volumes/SystemRoot/usr/local/lib/i386-linux-gnu/darling` in `$LD_LIBRARY_PATH` -- how do we do this without hardcoding i386 (we only build it for x64)?

LubosD:
> how do we do this without hardcoding i386 (we only build it for x64)?

Having both in doesn't do any harm...

[...]

PR #236, opened 2017-01-15
Fix OpenSSL libraries path

X0rg:
It refers to PR darlinghq/darling-openssl#1.
I was unable to install the darling-git package on Arch due to incorrect path for libssl098.so and libcrypto098.so.


ahyattdev:
Similar fixes are needed for i386 builds, no?

X0rg:
No: 32-bit build is installed in $PREFIX/usr/lib32, and that's right (on Arch).

X0rg:
@ahyattdev: thanks!

X0rg:
Also you need to merge darlinghq/darling-openssl#1 (see 51c8004). ;)

bugaevc:
> No: 32-bit build is installed in $PREFIX/lib32, and that's right (on Arch).

Well, it's actually wrong for Fedora (and probably other RH-related distros) -- we have `lib` and `lib64` here, so it should better be conditional just like you've made it for 64.

But, as mentioned in #231, this is all temporary, so it doesn't matter much.

bugaevc:
> Also you need to merge darlinghq/darling-openssl#1 (see 51c8004). ;)

Done

X0rg:
Yes, sadly, the FHS on Arch is absolutely not conformant (in fact, /usr/lib64 exists, but it is a symlink on /usr/lib for compatibility, so packages should not use /usr/lib64).

Issue #238, opened 2017-01-16
getaddrinfo() hangs


bugaevc:
To reproduce (may be not deterministic):
```
Darling [~]$ python
>>> from urllib2 import urlopen
>>> urlopen("http://example.com")

(hangs)
```

I've experienced this issue with other progs (e.g. native `git` from Rudix) as well.

Some debugging shows that `dns_res_send` calls `notify_register_plain`, which calls [`_notify_lib_self_state`](https://github.com/darlinghq/darling/blob/master/src/libnotify/notify_client.c#L937), which calls [`dispatch_once`](https://github.com/darlinghq/darling-libdispatch/blob/master/src/once.c#L36) with `val = &globals->self_state_once`. [`dispatch_once_f`](https://github.com/darlinghq/darling-libdispatch/blob/master/src/once.c#L46) then reinterprets `*val` (that is `globals->self_state_once` of `libnotify_client`) as a pointer, which should initially be set to `NULL` -- if it's not, it means some other thread is already executing this function, and `dispatch_once_f` waits on a semaphore for that thread to complete its work.

The problem here is that `globals->self_state_once` isn't `NULL`. The `globals` structure is allocated by [`_notify_globals`](https://github.com/darlinghq/darling/blob/master/src/libnotify/notify_internal.h#L92) function using either `calloc` or `os_alloc_once` (depending on availablity of `os_alloc_once`), and then initialized by [`_notify_init_globals`](https://github.com/darlinghq/darling/blob/master/src/libnotify/notify_client.c#L222), which, in fact, only initializes a few members. No other code ever touches `self_state_once`.

With `calloc`, however, the whole `globals` structure is zeroed out, meaning that `self_state_once` indeed gets the initial value of `NULL`. That's not the case with [`os_alloc_once`](https://github.com/darlinghq/darling/blob/master/src/libc/os/alloc_once_impl.c), which uses plain `malloc`.

I wasn't able to found any documentation about `os_alloc_once`, but since it's being used as an alternative to `calloc` here, I suppose it should zero out the allocated memory.

Issue #233, opened 2017-01-13
Segfault in os_alloc_once


bugaevc:
`libSystem_initializer` calls `__keymgr_initializer`, it eventually calls `_keymgr_globals`, which [calls](https://github.com/darlinghq/darling/blob/master/src/keymgr/keymgr.c#L160) [`os_alloc_once`](https://github.com/darlinghq/darling/blob/master/src/libc/os/alloc_once_impl.c) with `ctor = NULL`.  `os_alloc_once` calls `ctor` (the argument), and we get a segfault.

It doesn't seem like `__keymgr_initializer` gets called at all on x64 (is it conditionally compiled? at least `gdb` won't step into the call or break on the breakpoint), but on x86 this causes a segfault.

Since `keymgr.c` is Apple's code, and `alloc_once_impl.c` is ours, this should be probably fixed like:
```C
if (ctor != NULL)
    ctor(blocks[key]);
```

PR #239, opened 2017-01-16
Fix alloc_once implementation

bugaevc:
The proposed fixes for issues #233 and #238.


LubosD:
I'm surprised the memory returned by malloc() isn't already zeroed out, because that was one of the compatibility "differences" I noticed on macOS.

Issue #261, opened 2017-03-23
Arch build issues.


DuIslingr:
Initially I had an issue building due to a discrepancy with uint64 type in darling/platform-include/i386/types.h with the one on my system. I resolved it by changing it to the following

Line 46
```c
#ifndef __LP64__
typedef long long 		__int64_t;
typedef unsigned long long	__uint64_t;
#else
#undef __int64_t
#undef __uint64_t
typedef long int		__int64_t;
typedef unsigned long int	__uint64_t;
#endif /* !__LP64__ */
```

But now I get a new build error farther along which I can't seem to find any info about.
```bash
/home/duislingr/build/darling/src/libc/stdtime/FreeBSD/time32.c:47:1: error: 
      conflicting types for '_time64_to_time'
_time64_to_time(__int64_t t64)
^
/home/duislingr/build/darling/src/libc/include/timeconv.h:54:8: note: previous
      declaration is here
time_t _time64_to_time(int64_t t64);
       ^
/home/duislingr/build/darling/src/libc/stdtime/FreeBSD/time32.c:57:1: error: 
      conflicting types for '_time_to_time64'
_time_to_time64(time_t t)
^
/home/duislingr/build/darling/src/libc/include/timeconv.h:55:9: note: previous
      declaration is here
int64_t _time_to_time64(time_t t);
        ^
```

If its of any note the Arch system is Apricity OS and is running 4.10 kernel.

LubosD:
Could you please show us the original error, prior to modifying `i386/types.h`?

DuIslingr:
Hmmm odd i commented out the code I added and the error will not reproduce anymore :/ I even did make clean.

However this was the original error

```bash
[  1%] Built target migcom
[  1%] Built target system_duct
[  1%] Building C object src/kernel/emulation/linux/CMakeFiles/emulation.dir/misc/sysctl.o
In file included from /home/duislingr/build/darling/src/kernel/emulation/linux/misc/sysctl.c:6:
In file included from /home/duislingr/build/darling/platform-include/mach/host_info.h:68:
In file included from /home/duislingr/build/darling/platform-include/mach/message.h:75:
In file included from /home/duislingr/build/darling/platform-include/mach/port.h:92:
In file included from /home/duislingr/build/darling/platform-include/mach/machine/vm_types.h:33:
In file included from /home/duislingr/build/darling/platform-include/mach/i386/vm_types.h:72:
[...]

PR #268, opened 2017-03-28
Fixes for setup-ld-so.sh

X0rg:
Hello,

As mentioned in #261, the `$INSTALL_PREFIX/libexec/darling/etc/ld.so.conf.d` directory is empty when packaging Darling.

We shouldn't modify the _real_ filesystem directly. Files generated at build time should be put inside `build` directory instead of root directory.

I think it's better to pass working directory as argument in `setup-ld-so.sh`, this script should be run as regular user.
BTW, the [last line](https://github.com/darlinghq/darling/blob/c7a1b4a2e7fffe5d82f4af2ff06ef002ce98a2eb/src/setup-ld-so.sh#L24) remains broken. Maybe we can do that in `setupPrefix()` in `darling.c`, or something like that?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,7 @@ set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
 include(InstallSymlink)
 
 set(DARLING_TOP_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}")
+set(LD_SO_CONF_DIRECTORY "${CMAKE_BINARY_DIR}/etc/ld.so.conf.d")
 set(DARLING_NO_EXECUTABLES OFF)
 set(CMAKE_C_IMPLICIT_LINK_LIBRARIES "")
 set(CMAKE_CXX_IMPLICIT_LINK_LIBRARIES "")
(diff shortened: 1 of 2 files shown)


bugaevc:
> We shouldn't modify the real filesystem directly. Files generated at build time should be put inside build directory instead of root directory.

That's how it already is. The script in question is being run during `make install`, not `make`.

Besides, it won't easily work the way you propose. In particular, you'd need root because mount namespaces can only be created by root; you could, theoretically, workaround this by using user namespaces too.

What is the problem you're trying to solve in the first place? Why is that directory empty, and how is that related to the rest of this PR?

X0rg:
The `make install` is writing files in `$PREFIX/libexec/darling/etc/ld.so.conf.d` (where `PREFIX=/usr` when you do a package), that's modifying the _real_ root directly.

In 6cd989c, the goal is to write these files in a temporary directory (e.g. build directory), allowing to put Darling in a package (and redistribute it).
This is not perfect because it's based on build machine configuration, not on host configuration (I want to mean, an user who installs a package), but it solves the empty `$INSTALL_PREFIX/libexec/darling/etc/ld.so.conf.d` directory.

When you build a package, you don't have root permissions, the installation is performed in a fakeroot.
And then, this fakeroot is put inside a compressed archive.

bugaevc:
Let me put it this way:

You're trying to make an Arch package, and the package build system doesn't run `make install` as the real root (only fakeroot), so everything works except the unshare/mount/chroot thing in this script. That is why the directory ends up empty.

Is that correct? How come it worked for you before? We've been relying on ld-setup for several months now.

The way to fix this is to either alter the package build system to allow using the real root (but that's out of the question, I suppose), or use a user namespace to make us the "real" root inside that namespace.

In that case changes this PR actually introduces won't help and aren't needed; implement the namespace thing instead.

bugaevc:
Do you actually mean [`fakeroot(1)`](http://man.he.net/man1/fakeroot), or just a fake root directory? If it's the latter, then why doesn't it work, what causes the issue?

LubosD:
@bugaevc I suppose they use `make DESTDIR=xyz install`. This means `CMAKE_INSTALL_PREFIX` is still `/usr`, but installation isn't happening into `/`.

X0rg:
I'm not trying to make an Arch package, that's done since 2 years. I'm trying to fix things which assume you'll **run** Darling on the same OS that **build** Darling, but yes, everything works except work done by `setup-ld-so.sh` script.

> The way to fix this is to either alter the package build system to allow using the real root
What? You are scaring me. This is NOT a redistributable way.

> Do you actually mean fakeroot(1), or just a fake root directory?
`makepkg` use `fakeroot` internally, but that's not the question.

@LubosD: That's the case.

X0rg:
This is the list of files present in package: [files_list.txt](https://github.com/darlinghq/darling/files/878148/files_list.txt). As you can see on line 25, `ld.so.conf.d` is empty.

But I understand why you'll not merge this PR.
This script must be run on package installation (e.g. on host computer), not on package creation.
One solution is to add `setup-ld-so.sh` script into package: it can be automatically run as root on installation.

Wait and see...

X0rg:
Done in X0rg/AUR-PKGBUILD@5c2a8f86dce77bdeb85df7d072687f1b885a0587.

bugaevc:
@X0rg awesome!

PR #276, opened 2017-04-05
Fix running programs directly via `darling prog`

yan12125:
Fixed a problem like this:
```
$ darling /usr/local/libexec/darling/usr/bin/arch
Cannot open /Volumes/SystemRoot/usr/local/libexec/da���: No such file or directory
```
The memory is corrupted :)

### DIFF
--- a/src/dyld/mldr.c
+++ b/src/dyld/mldr.c
@@ -92,13 +92,18 @@ int main(int argc, char** argv, char** envp)
 	}
 
 	// sys_execve() passes the original file path appended to the mldr path in argv[0].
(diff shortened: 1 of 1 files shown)


bugaevc:
Thanks!

By the way, does `filename` ever get `free`d?

yan12125:
IIRC __builtin_alloca allocates memory on stack instead of on heap, so there's no need to free it :)

bugaevc:
Right. We could still free the memory, even if not with `free()`, but it doesn't matter much if the allocator is not involved. Thanks again!

PR #285, opened 2017-04-08
Add /bin/expr

yan12125:
Update after darlinghq/darling-shell_cmds#2


yan12125:
Thanks for such fast replies and the invitation!

LubosD:
Please join us on Slack as well. I've sent you an invitation.

PR #319, opened 2017-06-12
xcselect fixes for /usr/libexec/DeveloperTools, including all xcrun shims

stek29:
Add all shims (which were present on my macOS Sierra install).
Add REQUIRE_XCODE define support to xcrun-shim.c to specify flag for xcselect_invoke_xcrun

Also fix xcselect_invoke_xcrun invoking execv with invalid argv when executing from `/usr/libexec/DeveloperTools`

### DIFF
--- a/src/xcselect/CMakeLists.txt
+++ b/src/xcselect/CMakeLists.txt
@@ -16,20 +16,46 @@ target_link_libraries(xcrun system xcselect)
 add_darling_executable(xcode-select xcode-select.c)
 target_link_libraries(xcode-select system xcselect)
 
-function(add_shim name)
+install(TARGETS xcselect DESTINATION libexec/darling/usr/lib)
+install(TARGETS xcrun xcode-select DESTINATION libexec/darling/usr/bin)
+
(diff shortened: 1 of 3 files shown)


stek29:
Btw, What about changing 
`i < argc` to `i <= argc` on [line 246](https://github.com/darlinghq/darling/blob/f45df71132a8d5c98f383e1e34d1be4abe356dcb/src/xcselect/xcselect.c#L246) and `argc+2` to `argc+3` [on line 240](https://github.com/darlinghq/darling/blob/f45df71132a8d5c98f383e1e34d1be4abe356dcb/src/xcselect/xcselect.c#L240)?

Because if I understand correctly agrv2 is not `NULL` terminated currently.

stek29:
Also, it should probably check if file exists before executing it in `/usr/libexec/DeveloperTools`, and if it doesn't fall to "cannot find developer tools" error.

stek29:
Oh, wait, I've broken error messages...

stek29:
Ok, it should be fine now.

PR #320, opened 2017-06-12
Allow execve with empty argv

stek29:
Tested with following snippet which runs fine on Linux and macOS but causes `mldr is part of Darling. It is not to be executed directly.` on Darling.

```c
#include <unistd.h>
int main(void) {
  char *argv2[] = {NULL};
  execv("/bin/bash", argv2);
  return 1;
}
```
I haven't found any regressions, but please test it.

### DIFF
--- a/src/startup/mldr.c
+++ b/src/startup/mldr.c
@@ -91,17 +91,23 @@ int main(int argc, char** argv, char** envp)
 	void** sp;
 	int pushCount = 0;
 	const char* apple[3];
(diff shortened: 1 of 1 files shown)


bugaevc:
Thanks! Looks good, except your system's time may be somewhat off -- we have definitely discussed this less than 9 hours ago.

stek29:
@bugaevc Yup, just noticed it too, forgot to set up ntp after fresh install :)

Issue #338, opened 2017-08-22
can't build darling on fedora 26


Kidlike:
I followed the instructions here:

https://wiki.darlinghq.org/build_instructions

When I run `make`, it throws this error:
```
[ 95%] Built target elf
[ 95%] Built target dwarf
[ 95%] Built target ctf
[ 95%] Linking CXX shared library libdtrace.dylib
ld: warning: directory not found for option '-F/System/Library/Frameworks/'
ld: warning: -rpath cannot be used with dylibs that will be in the dyld shared cache
ld: warning: directory not found for option '-F/System/Library/Frameworks/'
ld: warning: -rpath cannot be used with dylibs that will be in the dyld shared cache
[ 96%] Built target dtrace
[ 96%] Building C object src/external/cocotron/CoreGraphics/CMakeFiles/CoreGraphics_obj.dir/CGLPixelSurface.o
In file included from /home/vagrant/install/src/darling/darling/src/external/cocotron/CoreGraphics/CGLPixelSurface.m:9:
/home/vagrant/install/src/darling/darling/src/external/cocotron/AppKit/O2Surface_cairo.h:9:9: fatal error: 'cairo.h' file not found
#import <cairo.h>
        ^~~~~~~~~
1 error generated.
make[2]: *** [src/external/cocotron/CoreGraphics/CMakeFiles/CoreGraphics_obj.dir/build.make:495: src/external/cocotron/CoreGraphics/CMakeFiles/CoreGraphics_obj.dir/CGLPixelSurface.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:48864: src/external/cocotron/CoreGraphics/CMakeFiles/CoreGraphics_obj.dir/all] Error 2
make: *** [Makefile:130: all] Error 2
```

This file exists in the following location though:

```
$ locate -r \/cairo.h$
/usr/include/cairo/cairo.h
```

Any ideas ?

Thanks

stek29:
> Any ideas ?

Well, it includes `cairo.h` and not `cairo/cairo.h`, and `/usr/include/cairo` isn't in include path.

bugaevc:
* It builds for me, or at least was building some time ago (I'll check when I get home)
* ` /usr/include/cairo` [should be](https://github.com/darlinghq/darling-cocotron/blob/master/CoreGraphics/CMakeLists.txt#L39) in the include path.

bugaevc:
Yes, it build just fine for me.

[...]

PR #372, opened 2018-01-23
WIP: Adding rpm for darling

andyneff:
Added an rpm for darling, separate rpm for kernel module (darling-mach) that uses dkms. This pretty much parallels the debian package.

Wrapped all up in a docker for "ease of use".

I used this for testing #338

Looking for some feedback


ahyattdev:
Thanks!

PR #404, opened 2018-07-19
Fedora 28

andyneff:
Updated to work with newest changes and on Fedora 28

- Updated package names that needed 64 bit specified on Fedora 28 for some reason
- Added new package requirements
- Removed adding mdlr capabilities
- Added OpenGL_GL_PREFERENCE=GLVND to CMAKE call, else it doesn't know whether to use legacy or glvnd
- Added copy of rtsig.h to kernel module source dir


rpm/build.bsh
```diff
@@ -7,7 +7,7 @@ if [ -e "/src/rpm/SOURCES/darling.tar.gz" ]; then
   ln -s /src/rpm/SOURCES/darling.tar.gz /root/rpmbuild/SOURCES/
 else
   # Preparing tarball
-  tar --transform "s|^\./|/darling/|" -cf /root/rpmbuild/SOURCES/darling.tar.gz -C /src --exclude=.git --exclude SOURCES .
+  tar --transform "s|^\./|./darling/|" -cf /root/rpmbuild/SOURCES/darling.tar.gz -C /src --exclude=.git --exclude SOURCES .
```
andyneff: Adding the . was more proper, and removed a warning

bugaevc:
Looks great, thanks!

PR #442, opened 2018-10-22
Changes to allow building on Fedora 28

woachk:
With these two changes, Darling becomes buildable with docker-compose for Fedora 28.
These changes aren't needed to build without using RPMs.


bugaevc:
Hi, thanks.

`libbsd-devel` dependency indeed makes sense. `udev/rules.d/00-darling-mach.rules` isn't needed anymore, we no longer use a userspace-accessible `/dev/mach` device, so should we just remove that line?

cc @andyneff

andyneff:
LGTM

PR #470, opened 2019-01-23
Update src/tools/sudo.c with more features

JL2210:
Signed-off-by: James Larrowe <larrowe.semaj11@gmail.com>

Compiles with `-Wall -Wextra -Werror`
Seemingly runs fine, do you have a testsuite?
Just noticed, I was looking at the manpage for Linux's `sudo`. I can update that, if you'd like.

### DIFF
--- a/src/tools/sudo.c
+++ b/src/tools/sudo.c
@@ -1,26 +1,68 @@
 #include <unistd.h>
 #include <stdio.h>
+#include <string.h>
 
-int main(int argc, char** argv)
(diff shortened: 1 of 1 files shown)


JL2210:
What are your thoughts on this? I have another branch ready with a more MacOS compliant sudo, if needs arise.
This is the page I'm referring to: [Sudo MacOS](https://ss64.com/osx/sudo.html)

LubosD:
It's merged already, so we're happy to accept your further PRs.

PR #505, opened 2019-05-04
Add basic DrawSprocket stub implementation.

jamesu:
Adds a stub for DrawSprocket with exported functions, including typed parameters in the unfortunate event that an application actually calls said functions.


ahyattdev:
Looks great to me. By the way, we have a generator for creating framework stubs located [here](https://github.com/darlinghq/darling/blob/master/tools/darling-stub-gen).

I'll write something on the wiki about how to use it

ahyattdev:
We now have a wiki page describing the process: https://wiki.darlinghq.org/generating_stubs

This should make it easier for you to create stubs!

PR #517, opened 2019-05-20
Implement CNLabeledValue

CuriousTommy:
I created stubs for the `CNLabeledValue` class. These functions are based on what the Apple developer website described. For the coding style, I went with how the website written the functions/variables. Hopefully I did not mess up here, but let me know if you want me to change anything.

With that being said, I do have a few questions/comments:

- I noticed that there is are two functions in the `CNLabeledValue.m`. I am not sure what they are used for so I left them there:

```
- (NSMethodSignature *)methodSignatureForSelector:(SEL)aSelector;
- (void)forwardInvocation:(NSInvocation *)anInvocation;
```

- For iOS and watchOS, the variables are `nonatomic` while MacOS is `atomic`. I recall there being a way to the let the compiler know if you want to use nonatomic or atomic, but I don't remember how to implement it (and my google skills are failing me).

- I ordered the functions and variables based on the website.

### DIFF
--- a/src/Contacts/src/CNLabeledValue.m
+++ b/src/Contacts/src/CNLabeledValue.m
@@ -22,35 +22,69 @@
 @implementation CNLabeledValue
 
 - (instancetype)initWithLabel:(NSString *)label value:(id)value {
+    if (self = [super init]) {
+//        [self applyLabeledValueVariablesIdentifier: [[NSUUID alloc] init].UUIDString label:label value:value];
+        [self applyLabeledValueVariablesIdentifier: [[NSUUID alloc] init].UUIDString label:label value:value];
+    }
+    
+    return self;
+}
+
+- (instancetype)initWithIdentifier:(NSString *)identifier label:(NSString *)label value:(id)value {
+    if (self = [super init]) {
+        [self applyLabeledValueVariablesIdentifier:identifier label:label value:value];
+    }
+    
+    return self;
+}
+
+- (void)dealloc {
+    [_identifier release];
+    [_label release];
+    [_value release];
+    [super dealloc];
+}
+
+- (instancetype)applyLabeledValueVariablesIdentifier:(NSString *)identifier label:(NSString *)label value:(id)value {
+    if (value == nil) {
+        @throw [[NSException alloc] initWithName:@"CNPropertyInvalidTypeExpression" reason:@"Variable value cannot be null." userInfo:nil];
+    }
+    
+    _identifier = [identifier copy];
+    _label = [label copy];
+    _value = [value copy];
     return self;
 }
 
 + (instancetype)labeledValueWithLabel:(NSString *)label value:(id)value {
-    id new_class = [[[self class] alloc] init];
-    return new_class;
+    return [[[[self class] alloc] initWithLabel:label value:value] autorelease];
 }
 
 - (instancetype)labeledValueBySettingLabel:(NSString *)label {
-    return self;
+    return [[[[self class] alloc] initWithIdentifier:_identifier label:label value:_value] autorelease];
 }
 
 - (instancetype)labeledValueBySettingLabel:(NSString *)label value:(id)value {
-    return self;
+    return [[[[self class] alloc] initWithIdentifier:_identifier label:label value:value] autorelease];
 }
 
 - (instancetype)labeledValueBySettingValue:(id)value {
-    return self;
+    return [[[[self class] alloc] initWithIdentifier:_identifier label:_label value:value] autorelease];
 }
 
+/*
+ TODO: Add support for obtaining localized string from contact label
+ */
 + (NSString *)localizedStringForLabel:(NSString *)label {
-    return @"";
+    // If you provide a custom label, it will return the same label.
+    return label;
 }
 
 
 // NSCopy
-- (nonnull id)copyWithZone:(nullable NSZone *)zone {
+- (id)copyWithZone:(NSZone *)zone {
     id copy = [[[self class] alloc] init];
-    return copy;
+    return [copy autorelease];
 }
 
 
@@ -59,14 +93,13 @@
     return YES;
 }
 
-- (void)encodeWithCoder:(nonnull NSCoder *)aCoder {
+- (void)encodeWithCoder:(NSCoder *)aCoder {
     return;
 }
 
-- (nullable instancetype)initWithCoder:(nonnull NSCoder *)aDecoder {
+- (instancetype)initWithCoder:(NSCoder *)aDecoder {
     return self;
 }
-
 
 - (NSMethodSignature *)methodSignatureForSelector:(SEL)aSelector {
     return [NSMethodSignature signatureWithObjCTypes: "v@:"];
--- a/src/Contacts/src/CNLabeledValue.m
+++ b/src/Contacts/src/CNLabeledValue.m
@@ -19,23 +19,21 @@
 
 #import <Contacts/CNLabeledValue.h>
 
+#define NSCODER_IDENTIFIER @"_identifier"
+#define NSCODER_LABEL @"_label"
+#define NSCODER_VALUE @"_value"
+
+const NSExceptionName CNPropertyInvalidTypeExceptionName = @"CNPropertyInvalidTypeException";
+
 @implementation CNLabeledValue
 
-- (instancetype)initWithLabel:(NSString *)label value:(id)value {
-    if (self = [super init]) {
-//        [self applyLabeledValueVariablesIdentifier: [[NSUUID alloc] init].UUIDString label:label value:value];
-        [self applyLabeledValueVariablesIdentifier: [[NSUUID alloc] init].UUIDString label:label value:value];
-    }
-    
-    return self;
-}
+@synthesize identifier = _identifier;
+@synthesize label = _label;
+@synthesize value = _value;
 
-- (instancetype)initWithIdentifier:(NSString *)identifier label:(NSString *)label value:(id)value {
-    if (self = [super init]) {
-        [self applyLabeledValueVariablesIdentifier:identifier label:label value:value];
-    }
-    
-    return self;
+- (instancetype)initWithLabel:(NSString *)label value:(id)value {
+    NSUUID *uuid = [NSUUID UUID];
+    return [self initWithIdentifier: [uuid UUIDString] label:label value:value];
 }
 
 - (void)dealloc {
@@ -45,31 +43,20 @@
     [super dealloc];
 }
 
-- (instancetype)applyLabeledValueVariablesIdentifier:(NSString *)identifier label:(NSString *)label value:(id)value {
-    if (value == nil) {
-        @throw [[NSException alloc] initWithName:@"CNPropertyInvalidTypeExpression" reason:@"Variable value cannot be null." userInfo:nil];
-    }
-    
-    _identifier = [identifier copy];
-    _label = [label copy];
-    _value = [value copy];
-    return self;
-}
-
 + (instancetype)labeledValueWithLabel:(NSString *)label value:(id)value {
-    return [[[[self class] alloc] initWithLabel:label value:value] autorelease];
+    return [[[self alloc] initWithLabel:label value:value] autorelease];
 }
 
 - (instancetype)labeledValueBySettingLabel:(NSString *)label {
-    return [[[[self class] alloc] initWithIdentifier:_identifier label:label value:_value] autorelease];
+    return [[[CNLabeledValue alloc] initWithIdentifier:_identifier label:label value:_value] autorelease];
 }
 
 - (instancetype)labeledValueBySettingLabel:(NSString *)label value:(id)value {
-    return [[[[self class] alloc] initWithIdentifier:_identifier label:label value:value] autorelease];
+    return [[[CNLabeledValue alloc] initWithIdentifier:_identifier label:label value:value] autorelease];
 }
 
 - (instancetype)labeledValueBySettingValue:(id)value {
-    return [[[[self class] alloc] initWithIdentifier:_identifier label:_label value:value] autorelease];
+    return [[[CNLabeledValue alloc] initWithIdentifier:_identifier label:_label value:value] autorelease];
 }
 
 /*
@@ -81,10 +68,27 @@
 }
 
 
-// NSCopy
+// Private method.
+- (instancetype)initWithIdentifier:(NSString *)identifier
+                             label:(NSString *)label
+                             value:(id)value {
+    if (self = [super init]) {
+        if (value == nil) {
+            [NSException raise:CNPropertyInvalidTypeExceptionName format:@"Variable value cannot be null."];
+        }
+        
+        _identifier = [identifier copy];
+        _label = [label copy];
+        _value = [value copy];
+    }
+    
+    return self;
+}
+
+// NSCopying
 - (id)copyWithZone:(NSZone *)zone {
-    id copy = [[[self class] alloc] init];
-    return [copy autorelease];
+    id copy = [[[self class] alloc] initWithIdentifier:_identifier label:_label value:_value];
+    return copy;
 }
 
 
@@ -93,13 +97,29 @@
     return YES;
 }
 
+/*
+ When you look at the original CNLabeledValue .plist, there are also
+ three other keynames: linkedIdentifiers, storeIdentifier, storeInfo.
+ 
+ TODO: Figure out what they are used for */
 - (void)encodeWithCoder:(NSCoder *)aCoder {
+    [aCoder encodeObject:_identifier forKey:NSCODER_IDENTIFIER];
+    [aCoder encodeObject:_label forKey:NSCODER_LABEL];
+    [aCoder encodeObject:_value forKey:NSCODER_VALUE];
+    
     return;
 }
 
 - (instancetype)initWithCoder:(NSCoder *)aDecoder {
-    return self;
+    NSString *temp_identifier = [aDecoder decodeObjectOfClass:[NSString class] forKey:NSCODER_IDENTIFIER];
+    NSString *temp_lable = [aDecoder decodeObjectOfClass:[NSString class] forKey:NSCODER_LABEL];
+    NSString *temp_value = [aDecoder decodeObjectOfClass:[NSString class] forKey:NSCODER_VALUE];
+    
+    // Since initWithIdentifier already uses copy for the variable, I decided just call
+    // that function instead.
+    return [self initWithIdentifier:temp_identifier label:temp_lable value:temp_value];
 }
+
 
 - (NSMethodSignature *)methodSignatureForSelector:(SEL)aSelector {
     return [NSMethodSignature signatureWithObjCTypes: "v@:"];
--- a/src/Contacts/src/CNLabeledValue.m
+++ b/src/Contacts/src/CNLabeledValue.m
@@ -21,6 +21,53 @@
 
 @implementation CNLabeledValue
 
+- (instancetype)initWithLabel:(NSString *)label value:(id)value {
+    return self;
+}
+
++ (instancetype)labeledValueWithLabel:(NSString *)label value:(id)value {
+    id new_class = [[[self class] alloc] init];
+    return new_class;
+}
+
+- (instancetype)labeledValueBySettingLabel:(NSString *)label {
+    return self;
+}
+
+- (instancetype)labeledValueBySettingLabel:(NSString *)label value:(id)value {
+    return self;
+}
+
+- (instancetype)labeledValueBySettingValue:(id)value {
+    return self;
+}
+
++ (NSString *)localizedStringForLabel:(NSString *)label {
+    return @"";
+}
+
+
+// NSCopy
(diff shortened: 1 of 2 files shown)


bugaevc:
Okay, honestly, I'm unsure what we should do here.

First of all, I'm really glad about the interest you're taking in Darling and in wanting to help — this is great! Don't let the rest of this comment discourage you.

But... these are indeed stubs that don't work (do the correct thing). `labeledValueWithLabel:` even leaks memory (you're supposed to `autorelease` the value — look up Cocoa memory management conventions, or ask me for more details).

Considering we already have dynamic `forwardInvocation:`-based stubs (the objc runtime calls them when a method implementation is not found, and we just log a message), I'm not sure there is any advantage to having this.

Or maybe there is — it's easier to write an actual implementation on top of these stubs, when/if anyone ever gets to that. I'll leave it to @ahyattdev to decide.

Again, please don't be discouraged by this, and do continue learning objc! Feel free to ask me for help/pointers on any topic.

bugaevc:
Or — you could actually implement `CNLabeledValue`; there's not that much functionality missing from your stubs! Then we'd definitely want to merge it.

CuriousTommy:
> First of all, I'm really glad about the interest you're taking in Darling and in wanting to help — this is great! Don't let the rest of this comment discourage you.

Don't worry, I totally understand.

> Or maybe there is — it's easier to write an actual implementation on top of these stubs, when/if anyone ever gets to that. I'll leave it to @ahyattdev to decide.

> you could actually implement CNLabeledValue;

That's my original plan. I currently have a WIP implementation, but I wanted to make a stub so that I can revert my changes (in case if I do anything destructive, if that is even possible).

Should I close this PR and submit another one with most (if not all) of the functions implemented, or should I just implement it in this PR.

> But... these are indeed stubs that don't work (do the correct thing). labeledValueWithLabel: even leaks memory (you're supposed to autorelease the value — look up Cocoa memory management conventions, or ask me for more details).

Okay, I'll read up more on Cocoa's memory management. While I do my implementation, are there any other concerns I should keep in mind?

> Considering we already have dynamic forwardInvocation:-based stubs (the objc runtime calls them when a method implementation is not found, and we just log a message), I'm not sure there is any advantage to having this.

Ah so that is what `forwardInvocation` is used for. That's really cool.

bugaevc:
> I currently have a WIP implementation, but I wanted to make a stub so that I can revert my changes (in case if I do anything destructive, if that is even possible).

I don't really understand what you mean...

> Should I close this PR and submit another one with most (if not all) of the functions implemented, or should I just implement it in this PR.

Just push it here (maybe even amend and force-push), no need for a separate PR.

> Okay, I'll read up more on Cocoa's memory management. While I do my implementation, are there any other concerns I should keep in mind?

A word of caution: there are basically two different conventions, or systems (it's actually more than just a different convention) about Cocoa memory management: a newer one, ARC (automatic reference counting), and an older one (sometimes called MRC, for manual reference counting). You need MRC, not ARC — the one where you explicitly do `retain`s, `release`s and `autorelease`s.

> For iOS and watchOS, the variables are nonatomic while MacOS is atomic. I recall there being a way to the let the compiler know if you want to use nonatomic or atomic, but I don't remember how to implement it (and my google skills are failing me).

You let the compiler know whether you want atomic or nonatomic by saying it inside the parens in a `@property` declaration, like so:

```objc
@property (atomic) NSRunLoop *loop;
@property (nonatomic, weak) IBOutlet UILabel *label;
```
and you have already done that.

To get the atomic on macOS, nonatomic on iOS (and others) behavior, there's a `NS_NONATOMIC_IOSONLY` macro that (I believe) expands either to `atomic` or `nonatomic` depending on that platform. Or at least it should do that; it seems [the one we have](https://github.com/darlinghq/darling-foundation/blob/master/include/Foundation/NSObjCRuntime.h#L49) always expands to `nonatomic` (understandable given Apportable only cared about iOS apps) — it'd be nice if you fix that as well.

CuriousTommy:
> I don't really understand what you mean...

Having multiple commits is better then having one commit when you want to `git revert` on a commit that breaks Darling.

But the way I worded it is really bad... you can just ignore that message...

> You need MRC, not ARC — the one where you explicitly do retains, releases and autoreleases.

Good to know, I'll keep that in mind.

> To get the atomic on macOS, nonatomic on iOS (and others) behavior, there's a NS_NONATOMIC_IOSONLY macro that (I believe) expands either to atomic or nonatomic depending on that platform.

Oh! That could work. My only concern is if this would make watchOS `atomic` instead of `nonatomic`?

> Or at least it should do that; it seems the one we have always expands to nonatomic (understandable given Apportable only cared about iOS apps) — it'd be nice if you fix that as well.

Outside of importing headers or `#ifnder CLASS_H #define CLASS_H ... #endif`, I never really messed around with the preprocessor. But if it is not too hard to implement, I can it give it a try.

bugaevc:
> Having multiple commits is better then having one commit when you want to `git revert` on a commit that breaks Darling.

Sure, it usually makes sense to separate changes into (somewhat) independent, self-contained commits. But in the case, it doesn't make that much sense to separate adding stubs and implementing them. Even if you do want to separate them, there's no need to "land" the changes as separate PRs — you would usually do (and commit) both parts and then create a PR with both.

> Oh! That could work. My only concern is if this would make watchOS atomic instead of nonatomic?

I *suspect* (I don't know for sure, this is something for you to maybe check!) that the name `NS_NONATOMIC_IOSONLY` actually predates watchOS, and that it actually means "atomic on macOS only", which is what we need.

> But if it is not too hard to implement, I can it give it a try.

It shouldn't be too hard, it's something like

```objc
#ifndef NS_NONATOMIC_IOSONLY
#if TARGET_OS_MAC
#define NS_NONATOMIC_IOSONLY atomic
#else
#define NS_NONATOMIC_IOSONLY nonatomic
#endif
#endif
```

but make sure to look up & use the appropriate macro in place of `TARGET_OS_MAC` (one that is 1 on macOS but 0 on iOS/watchOS).

> After seeing a video on MRC, I think I understand the basics.

I can also recommend this post: https://www.tomdalling.com/blog/cocoa/an-in-depth-look-at-manual-memory-management-in-objective-c/

> Since I am using `copy` instead of `retain` in my `property`, do I still need to worry about running `[_variable retain]`?

If you're using the property via accessors (e.g. `self.foo = bar`, or `[self setFoo: bar]`/`[self foo]` — you can do that from objects own implementation, not just from the outside!), you don't to worry about this, cause the autogenerated accessor will do the right thing automatically.

But it's recommended not to use property accessors in `init...` and `dealloc` methods, so there you'll have to access the ivars directly.

For example:
```objc
- (instancetype) initWithFoo: (Foo *) foo {
    self = [super init];
    if (self == nil) {
        return nil;
    }

    _foo = [foo copy];

    return self;
}
```

Why are there no retains/releases here? That's because we don't keep a reference to the *original* `foo` we were passed — we just make a copy of it and then forget about it. We don't we retain the copy since we do keep a reference to it in our ivar? That's because the result of `copy` is "already retained" (as are results of `alloc`, `init...` and `new...` methods).

That is the convention — if a method is one of these, it returns an already retained reference, and the caller doesn't have to retain it again (but it of course still has to release it later). The way the callee — in this case, `copy` — does that is perhaps by creating a new object, the copy, with `[[Foo alloc] init]` and _not_ autoreleasing it. Whereas "normal" methods that don't return an already retained reference either return another object (their ivar, or a result of some other method, or a string literal...), or create a new object and call `autorelease` on it before returning it.

(Hope that was clear, tell me if it's not!)

src/Contacts/src/CNLabeledValue.m
```diff
@@ -22,35 +22,69 @@
...
+- (void)dealloc {
+    [_identifier release];
+    [_label release];
+    [_value release];
+    [super dealloc];
+}
+
+- (instancetype)applyLabeledValueVariablesIdentifier:(NSString *)identifier label:(NSString *)label value:(id)value {
+    if (value == nil) {
+        @throw [[NSException alloc] initWithName:@"CNPropertyInvalidTypeExpression" reason:@"Variable value cannot be null." userInfo:nil];
```
bugaevc: Use [`[NSException raise:format:]`](https://developer.apple.com/documentation/foundation/nsexception/1569524-raise?language=objc)

Is "CNPropertyInvalidTypeExpression" the name that Apple's implementation raises?

src/Contacts/src/CNLabeledValue.m
```diff
@@ -22,35 +22,69 @@
...
+}
+
+- (void)dealloc {
+    [_identifier release];
+    [_label release];
+    [_value release];
+    [super dealloc];
+}
+
+- (instancetype)applyLabeledValueVariablesIdentifier:(NSString *)identifier label:(NSString *)label value:(id)value {
```
bugaevc: There's no need for a separate method, it would be better to just put the code of this method into `initWithIdentifier:label:value:` and make `initWithLabel:value:` call that, like this:

```objc
- (instancetype) initWithLabel: (NSString *) label value: (id) value {
    NSUUID *uuid = [NSUUID UUID];
    return [self initWithIdentifier: [uuid UUIDString] label: label value: value];
}

- (instancetype) initWithIdentifier: (NSString *) identifier
                              label: (NSString *) label
                              value: (id) value {

    if (self = [super init]) {
        _identifier = [identifier copy];
        // ...
    }
    return self;
}
```

src/Contacts/src/CNLabeledValue.m
```diff
@@ -22,35 +22,69 @@
 @implementation CNLabeledValue
 
 - (instancetype)initWithLabel:(NSString *)label value:(id)value {
+    if (self = [super init]) {
+//        [self applyLabeledValueVariablesIdentifier: [[NSUUID alloc] init].UUIDString label:label value:value];
+        [self applyLabeledValueVariablesIdentifier: [[NSUUID alloc] init].UUIDString label:label value:value];
```
bugaevc: You're leaking the UUID here. `alloc` returns you a retained object, and you're never releasing it.

Compare:
* `[[NSUUID alloc] init]` — `alloc` retains a retained instance, you own it and have to `release` it later
* `[NSUUID new]` — same
* `[[[NSUUID alloc] init] autorelease]` — properly released
* `[NSUUID UUID]` — returns unretained (as most methods do), and [the implementation](https://github.com/darlinghq/darling-foundation/blob/master/src/NSUUID.m#L36-L39) does `alloc`-`init`-`autorelease`

src/Contacts/src/CNLabeledValue.m
```diff
@@ -22,35 +22,69 @@
...
+    _identifier = [identifier copy];
+    _label = [label copy];
+    _value = [value copy];
     return self;
 }
 
 + (instancetype)labeledValueWithLabel:(NSString *)label value:(id)value {
-    id new_class = [[[self class] alloc] init];
-    return new_class;
+    return [[[[self class] alloc] initWithLabel:label value:value] autorelease];
```
bugaevc: There's no need to use `[self class]` here (and below), `self` is already the class (since this is a class method).

`[foo class]` is used for two reasons:
* To get a dynamic class of some _instace_, e.g. `[button class]` will likely evaluate to `NSButton`
* As a sort of "class literal": `[NSButton class]` will _also_ evaluate to `NSButton` (not to the metaclass!)

Why use `[NSButton class]` when you can say `NSButton` directly? That's an objc syntax quirk: you _cannot_ actually use `NSButton` (or other class names) directly in some contexts. You can _message_ it: `[NSButton foo]`, but you cannot e.g. store it in a variable or pass it as an argument:

```objc
Class buttonClass = NSButton;  // Nope, syntax error
[foo someMethodWithClass: NSButton];  // Nope

Class buttonClass = [NSButton class];  // Works
[foo someMethodWithClass: [NSButton class]];  // Works
```

This works because you are allowed to message the class, and `+ class` returns itself (as opposed to `- class`, the instance method, which returns the class of the instance).

Here, `self` is already a class and you can use it directly.

src/Contacts/src/CNLabeledValue.m
```diff
@@ -22,35 +22,69 @@
...
+ TODO: Add support for obtaining localized string from contact label
+ */
 + (NSString *)localizedStringForLabel:(NSString *)label {
-    return @"";
+    // If you provide a custom label, it will return the same label.
+    return label;
 }
 
 
 // NSCopy
```
bugaevc: There's no such thing as `NSCopy`, perhaps you mean `NSCopying`

src/Contacts/src/CNLabeledValue.m
```diff
@@ -22,35 +22,69 @@
...
-- (nonnull id)copyWithZone:(nullable NSZone *)zone {
+- (id)copyWithZone:(NSZone *)zone {
     id copy = [[[self class] alloc] init];
```
bugaevc: This will create a new empty instance, not a copy of `self`.

Also note that you cannot really call plain `init` if your class has custom inits (and does not override the plain `init`). It will compile, but it will fail at runtime much later when you're expecting something to be initialized and it's not.

src/Contacts/src/CNLabeledValue.m
```diff
@@ -22,35 +22,69 @@
...
-    return copy;
+    return [copy autorelease];
```
bugaevc: `copy`, in fact, should return a retained value, so there's no need for this `autorelease`...

CuriousTommy:
Thanks for look over my code, I will take a look at your comments and fix them. With that being said, I do have a few questions/statements.

I seem to get a compile error when I try to build my code ([link to full build log](https://pastebin.com/UQdh7MBU)) on Darling. The weird thing is that this code works perfectly fine on XCode.

I think the warning below gives an hint as to why it is not building. It says that I need to use `@synthesize`, but I thought that is no longer required (the compiler will do it for you automatically).
```
/home/user/Downloads/darling/src/Contacts/src/CNLabeledValue.m:22:17: warning: property 'label' requires method 'label' to be defined - use @synthesize, @dynamic or provide a method implementation in this class implementation [-Wobjc-property-implementation]
```
---
> I suspect (I don't know for sure, this is something for you to maybe check!) that the name NS_NONATOMIC_IOSONLY actually predates watchOS, and that it actually means "atomic on macOS only", which is what we need.

I wonder if there is a way for Objective-C to directly tell you if a variable is atomic or not. It would make it easy for me to do some testing on the simulator.

> (Hope that was clear, tell me if it's not!)

It took me a while to understand, but I think I sort-of understand what you are saying.

src/Contacts/src/CNLabeledValue.m
CuriousTommy: Yes, on MacOS you will get this exception message if you don't catch NSException:

```
*** Terminating app due to uncaught exception 'CNPropertyInvalidTypeException', reason: 'Labeled values require a nonnull value, but a null value was presented.'
```

src/Contacts/src/CNLabeledValue.m
CuriousTommy: So I changed `[[self class] alloc]` to `[self alloc]`. 

I don't get any warning when i do it to a class method (+), but if it were to do it to an instance method (-), I get a warning that `Instance method '-alloc' not found`.

bugaevc:
> I seem to get a compile error when I try to build my code (link to full build log) on Darling. The weird thing is that this code works perfectly fine on XCode.

> I think the warning below gives an hint as to why it is not building. It says that I need to use @synthesize, but I thought that is no longer required (the compiler will do it for you automatically).

I... actually don't know! I also thought Clang does it by default now. Some googling around reveals that there's an `-fobjc-default-synthesize-properties` flag that should've been turned on by default at some point, not sure what happened.

Perhaps you should just write the `@synthesize`s out explicitly.

> I wonder if there is a way for Objective-C to directly tell you if a variable is atomic or not. It would make it easy for me to do some testing on the simulator.

You can inspect a property using reflection (at runtime) using `class_getProperty()` and then `property_copyAttributeValue()`, or you could disassemble the generated setter and look for signs of locking.

But there are much easier ways. You can look at the preprocessed code to see what `NS_NONATOMIC_IOSONLY` expands to, or you can just look up the definition of `NS_NONATOMIC_IOSONLY` in Apple's header (but note that you cannot directly copy (non-free parts of) Apple code to Darling).

src/Contacts/src/CNLabeledValue.m
bugaevc: Does it look like Apple has a declaration like

```objc
const NSExceptionName CNPropertyInvalidTypeException = @"CNPropertyInvalidTypeException";
```

i.e. a symbol with the same name — or do they use a string literal directly? To check, you can run `nm` on the framework binary file.

src/Contacts/src/CNLabeledValue.m
bugaevc: Of course; my point about `[self class]` not being needed only applies to class methods, because there `self` is the class itself; in an instance method you need to use either `[self class]` (or `object_getClass(self)`, or even just `isa` — but don't do that) or `CNLabeledValue` (or `[CNLabeledValue class]` if you need it for something else than sending it a message).

The difference between `[self class]` and `CNLabeledValue` in instance methods, of course, is that `[self class]` is the *dynamic* class of the object. If `self` is an instance of some subclass of `CNLabeledValue`, `[self class]` will be that subclass.

src/Contacts/src/CNLabeledValue.m
CuriousTommy: Do you know where Apple keeps their framework binary files?

src/Contacts/src/CNLabeledValue.m
CuriousTommy: Oh, I see you mean now.

So if the Apple documentation tells me that `labeledValueBySettingLabel:` "returns a labeled value object", should I be using `[CNLabeledValue alloc]` instead of `[[self class] alloc]` (for a instance method)?

src/Contacts/src/CNLabeledValue.m
CuriousTommy: Yeah... I didn't really focusing on implementing NSCopying or NSSecureCoding. I wanted to focused on the main functions first before I implement these properly.

I will definitely keep your comments in mind when I actually implement these properties.

src/Contacts/src/CNLabeledValue.m
bugaevc: `/System/Library/Frameworks/Foo.framework/Foo`

src/Contacts/src/CNLabeledValue.m
bugaevc: I think so, yes.

In this case it's not that important since no one will probably subclass `CNLabeledValue` anyway (KVOA will, if you know what that is, but it would hide that by making `[self class]` return the original `CNLabeledValue` as if nothing happened).

src/Contacts/src/CNLabeledValue.m
CuriousTommy: [Here is the entire output if you want to look at it](https://pastebin.com/ftDwsi2V).

The closest name I could find is `_CNPropertyInvalidTypeExceptionName`.

bugaevc: Yeah, that sounds like it. The underscore is just Darwin name mangling, which means it's probably defined as

```objc
const NSExceptionName CNPropertyInvalidTypeExceptionName = @"CNPropertyInvalidTypeException";
```

CuriousTommy: Where should I put it? It doesn't seem like you are able to publicly access this variable (XCode complains that it is a undeclared variable).

bugaevc: That simply means it isn't declared in the header (only defined in the .m file)

CuriousTommy:
I updated my code. Hopefully I did things right this time. I haven't look into the `NS_NONATOMIC_IOSONLY` stuff yet.

> Perhaps you should just write the @synthesizes out explicitly.

After the `@synthesize` fix, I get another error...
```
/home/user/Downloads/darling/src/Contacts/src/CNLabeledValue.m:30:13: error: synthesized property 'identifier' must either be named the same as a compatible instance variable or must explicitly name an instance variable
@synthesize identifier = _identifier;
            ^
/home/user/Downloads/darling/src/Contacts/src/CNLabeledValue.m:31:13: error: synthesized property 'label' must either be named the same as a compatible instance variable or must explicitly name an instance variable
@synthesize label = _label;
            ^
/home/user/Downloads/darling/src/Contacts/src/CNLabeledValue.m:32:13: error: synthesized property 'value' must either be named the same as a compatible instance variable or must explicitly name an instance variable
@synthesize value = _value;
            ^
3 errors generated.
make[2]: *** [src/Contacts/CMakeFiles/Contacts.dir/build.make:297: src/Contacts/CMakeFiles/Contacts.dir/src/CNLabeledValue.o] Error 1
```
If I understand this error correctly, the `foo` from `@synthesize foo = _foo` should also be stated in `@property(..) MyClass *foo`, right? Or does it mean something else?

---
For the `initWithCoder`, the `[aDecoder decodeObjectOfClass:[NSString class]...]` would return a NSString object with `autorelease` applied to it, right?

CuriousTommy:
I am going to close this for now. In the future I'll look into this.

Issue #514, opened 2019-05-11
clock_sleep_trap_impl not implemented


jamesu:
`clock_sleep_trap_impl` is not implemented. 

This I assume is used internally by `usleep` and the like, which tends to be used a lot in games as well as threaded applications.

bugaevc:
> This I assume is used internally by `usleep` and the like

Actually it's not, they all work though `semaphore_timedwait[_signal]()`, though `usleep()` seems broken for some other reason, let me check...

jamesu:
My bad, might have to do a proper backtrace to see whats actually calling it in my case then - I assumed since it was spamming the console it had to be related to sleep.

bugaevc:
> though `usleep()` seems broken for some other reason, let me check...

no, it works, that was just an lldb bug

jamesu:
> was just an lldb bug

Is it just me or is lldb insanely buggy running in darling? Hope it improves as the project progresses ;)

bugaevc:
Not insanely, but somewhat buggy, yes. It's a complex piece of software that uses many internal Darwin details/APIs — it's a small miracle that it works at all, really, and we all should be eternally grateful to @LubosD for that :smile: Debugging without LLDB was way, way more complicated and slow.

jamesu:
After looking a little deeper, it seems `mach_wait_until` uses `clock_sleep`. This is called from `usleep` in this particular case, for example:

```
(lldb) bt
* thread #2: tid = 0x002f, 0xf752f0a8 libsystem_kernel.dylib`mach_wait_until, name = 'Torque Demo Deb', stop reason = breakpoint 1.1
  * frame #0: 0xf752f0a8 libsystem_kernel.dylib`mach_wait_until
    frame #1: 0xf76d0a14 libsystem_c.dylib`_nanosleep(requested_time=0xf35d8d68, remaining_time=0x00000000) + 548 at nanosleep.c:281
    frame #2: 0xf76cecd8 libsystem_c.dylib`_usleep(useconds=500000) + 72 at usleep.c:52
    frame #3: 0x00312d47 Torque Demo Debug OSX`Platform::sleep(unsigned int) + 23
    frame #4: 0x00300dac Torque Demo Debug OSX`NetAsync::run() + 396
    frame #5: 0x00300e04 Torque Demo Debug OSX`StartThreadFunc(int) + 36
    frame #6: 0x00311c55 Torque Demo Debug OSX`Thread::run(int) + 37
[...]

Issue #504, opened 2019-05-03
TGE based games: Missing Function stubs in Carbon & ApplicationServices


jamesu:
The following functions/symbols don't appear to be present, thus TGE-based games such as "Marble Blast" will not currently get past dyld when binding these frameworks. In this particular case, having stubs implemented would probably be sufficient enough to run in "headless" mode (i.e. no gui).

/System/Library/Frameworks/Carbon.framework/Versions/A/Carbon

   _ActivateTSMDocument
   _DeactivateTSMDocument
   _CallNextEventHandler
   _CreateEvent
   _CreateStandardAlert
   _CreateWindowGroup
   _FlushEvents
   _GetApplicationEventTarget
   _GetEventClass
   _GetEventDispatcherTarget
   _GetEventKind
   _GetEventParameter
   _GetMainEventQueue
   _GetScrapByName
   _GetScrapFlavorData
   _GetScrapFlavorSize
   _GetStandardAlertDefaultParams
   _GetWindowBounds
   _GetWindowEventTarget
   _GetWindowGroupLevel
   _GetWindowGroupOfClass
   _GetWindowPort
   _GetWindowPortBounds
   _HideMenuBar
   _InstallEventHandler
   _IsValidWindowPtr
   _NMInstall
   _NewEventHandlerUPP
   _NewTSMDocument
   _PostEventToQueue
   _ProcessHICommand
   _PutScrapFlavor
   _ReceiveNextEvent
   _ReleaseEvent
   _ReleaseWindow
   _RemoveEventHandler
   _RunApplicationEventLoop
   _RunStandardAlert
   _SelectWindow
   _SendEventToEventTarget
   _SetEventMask
   _SetEventParameter
   _SetWindowContentColor
   _SetWindowGroup
   _SetWindowGroupLevel
   _SetWindowTitleWithCFString
   _ShowMenuBar
   _TransitionWindowWithOptions
   _UseInputWindow
   _GetKeys

/System/Library/Frameworks/ApplicationServices.framework/Versions/A/ApplicationServices

   _ATSFontFindFromName
   _ATSFontGetHorizontalMetrics
   _ATSUCreateStyle
   _ATSUCreateTextLayout
   _ATSUDisposeStyle
   _ATSUDisposeTextLayout
   _ATSUDrawText
   _ATSUGetUnjustifiedBounds
   _ATSUMeasureTextImage
   _ATSUSetAttributes
   _ATSUSetLayoutControls
   _ATSUSetRunStyle
[...]

PR #523, opened 2019-06-04
Add more basic Carbon, ApplicationServices, CoreServices stubs

jamesu:
Adds more function stubs with correctly sized parameters to at least get torque-based games loading, satisfying #504 . Note that while this allows them to load, they currently do not function correctly due to #514 , threading issues, and (if one ever wanted to properly play them) there is no carbon ui or gl implemented.

Function prototypes have been added to relevant headers in carbon, though for now all the actual stubs are in Carbon.c and the existing MacWindows.cpp which has now also been added to the Carbon framework build.

### DIFF
--- a/src/ApplicationServices/ApplicationServices.c
+++ b/src/ApplicationServices/ApplicationServices.c
@@ -11,4 +11,319 @@ const CFStringRef kAXUIElementCopyHierarchyTruncateStringsKey = CFSTR("AXTRUNC")
 /* Not const for some reason */
 CFStringRef kAXTrustedCheckOptionPrompt = CFSTR("AXTrustedCheckOptionPrompt");
 
+#include <CoreFoundation/CoreFoundation.h>
+#include <CoreGraphics/CGDirectDisplay.h>
+#include <CoreGraphics/CGGeometry.h>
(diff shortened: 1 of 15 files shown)


ahyattdev:
Looks good to me, nice to see some Carbon stuff added. One thing to keep in mind is that if when running a program it says it expects a symbol in a framework it doesn't necessarily mean that the symbol is implemented in that framework. It is common for dynamic libraries to "reexport" others which is where they take all the exported symbols from another library and present them as their own.

In this case, ApplicationServices reexports CoreGraphics and that is why you got errors that those symbols in starting with CG were expected there.

What I've found useful is using MachOView to get a nice list of which libraries a dylib reexports.

Screen Shot 2019-06-05 at 8 36 17 PM

Those CoreGraphics symbols from what I've seen are also not present in our CoreGraphics, so those stubs you made can be moved over to CoreGraphics. I'll merge this and take care of that and set up ApplicationServices to reexport the things in the screenshot. It's now pretty easy since we just got a `reexport` CMake function.

Thanks for the PR!

ahyattdev:
I ran into an issue, will take longer than I thought to fix. Will be pushed soon.

PR #529, opened 2019-06-11
Add some AddressBook and Carbon stubs

kamillys:


ahyattdev:
Nice

src/Carbon/include/HIToolbox/TextInputSources.h
```diff
@@ -0,0 +1,19 @@
+#ifndef _Carbon_TextInputSources_H_
+#define _Carbon_TextInputSources_H_
+
+#ifndef __cplusplus
```
CuriousTommy: I don't have any experience with `extern "C"`, but going off of [stack overflow](https://stackoverflow.com/a/16192947/5988706), it should be `#ifdef __cplusplus` not `#ifndef __cplusplus`

However, I haven't tested this, so I could be wrong.

src/Carbon/include/HIToolbox/TextInputSources.h
```diff
@@ -0,0 +1,19 @@
...
+#endif
+
+extern const CFStringRef kTISPropertyInputSourceID;
+extern const CFStringRef kTISPropertyUnicodeKeyLayoutData;
+extern const CFStringRef kTISCategoryKeyboardInputSource;
+extern const CFStringRef kTISPropertyInputSourceCategory;
+extern const CFStringRef kTISPropertyInputSourceType;
+extern const CFStringRef kTISTypeKeyboardLayout;
+
+#ifndef __cplusplus
```
CuriousTommy: Same goes here.

src/Carbon/include/HIToolbox/TextInputSources.h
CuriousTommy: I don't have any experience with `extern "C"`, but going off of [stack overflow](https://stackoverflow.com/a/16192947/5988706), it should be `#ifdef __cplusplus` not `#ifndef __cplusplus`

However, I haven't tested this, so I could be wrong.

src/Carbon/include/HIToolbox/TextInputSources.h
CuriousTommy: Same goes here.

src/Carbon/include/HIToolbox/TextInputSources.h
kamillys: True, my bad.

PR #555, opened 2019-08-28
cmake: make sure the correct linker is used

### LINKED ISSUE #554: [CMake] Wrong linker being used for Mach-O cross compiling
I'm trying to build Darling on my Gentoo Linux. Clang is my default system compiler with custom patches. LLVM/LLD is my default ELF linker.

During building where it reaches the step it cross-compiles for Mach-O I noticed the wrong linker is being used resulting in funny error messages like this:

```plain
ld.lld: error: unknown argument '-dynamic', did you mean '-Bdynamic'
ld.lld: error: unknown argument '-dylib'
ld.lld: error: unknown argument '-arch'
ld.lld: error: unknown argument '-no_arch_warnings'
ld.lld: error: unknown argument '-dylib_install_name'
ld.lld: error: unknown argument '-compatibility_version'
ld.lld: error: unknown argument '-Z'
ld.lld: error: unknown argument '-dylib_file'
ld.lld: error: unknown argument '-dylib_file'
ld.lld: error: unknown argument '-dylib_file'
ld.lld: error: unknown argument '-dylib_file'
ld.lld: error: unknown argument '-dylib_file'
ld.lld: error: unknown argument '-dylib_file'
ld.lld: error: unknown argument '-dylib_file'
ld.lld: error: unknown argument '-dylib_file'
ld.lld: error: unknown argument '-dylib_file'
ld.lld: error: unknown argument '-dylib_file'
ld.lld: error: unknown argument '-dylib_file'
ld.lld: error: unknown argument '-dylib_file'
ld.lld: error: unknown argument '-dylib_file'
ld.lld: error: too many errors emitted, stopping now (use -error-limit=0 to see all errors)
```

I first was confused, but I figured out that the wrong linker is being invoked because I found out that the Apple ld was compiled earlier. I can workaround this problem my adding

```
set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fuse-ld=${CMAKE_BINARY_DIR}/src/external/cctools-port/cctools/ld64/src/x86_64-apple-darwin11-ld")
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fuse-ld=${CMAKE_BINARY_DIR}/src/external/cctools-port/cctools/ld64/src/x86_64-apple-darwin11-ld")
```

to every CMake file which cross compiles for Mach-O, but that's pretty annoying.

Is this a bug or a incompatibility with my custom toolchain. I don't have GCC installed. My default "binutils" are the LLVM binutils.

```
$ clang -v
clang version 9.0.0 
Target: x86_64-pc-linux
Thread model: posix
InstalledDir: /usr/lib/llvm/9/bin
Found CUDA installation: /opt/cuda, version 10.1
```

My clang defaults to libc++ and compiler-rt.

magiruuvelvet:
I figured out how to make CMake use the correct linker for all the cross compile targets. This patch fixes the issue I opened earlier.

fixes #554

### DIFF
--- a/cmake/use_ld64.cmake
+++ b/cmake/use_ld64.cmake
@@ -1,5 +1,10 @@
 FUNCTION(use_ld64 target)
 	set_property(TARGET ${target} APPEND_STRING PROPERTY
+		COMPILE_FLAGS " -fuse-ld=${CMAKE_BINARY_DIR}/src/external/cctools-port/cctools/ld64/src/x86_64-apple-darwin11-ld ")
+	set_property(TARGET ${target} APPEND_STRING PROPERTY
+		LINK_FLAGS " -fuse-ld=${CMAKE_BINARY_DIR}/src/external/cctools-port/cctools/ld64/src/x86_64-apple-darwin11-ld ")
+
+	set_property(TARGET ${target} APPEND_STRING PROPERTY
 		LINK_FLAGS " -B ${CMAKE_BINARY_DIR}/src/external/cctools-port/cctools/ld64/src/ \
 -B ${CMAKE_BINARY_DIR}/src/external/cctools-port/cctools/misc/ \
 -target x86_64-apple-darwin11 -Wl,-Z \


ahyattdev:
I'd like to understand the issue more before having this merged in. What is the cause? I don't think it's CMake related because I have successful builds with CMake 3.15.2 as well (but with Clang 8). Is it Gentoo related?

Has anyone else with Clang 9 had this issue?

Sounds like you have a very customized setup and I'd like to know this affects everyone with Clang 9 before merging this in.

magiruuvelvet:
Well looking at the CMake file I didn't spot a linker overwrite flag (`-fuse-ld=`) which is required to set a custom linker. Without this flag the compiler uses the default linker it was configured to use. With clang you can do that at the CMake stage when compiling it from source. My clang defaults to `ld.lld` (first result found in `$PATH`) when the linker isn't overwritten. I configured this myself. Clang defaults to just `ld` otherwise. Maybe that's the culprint?

What I found funny too is that the LLD linker has Mach-O support by invoking it with `ld64` as `argv[0]`. LLD supports many platforms and formats depending on how it was called.

What I'm trying to understand now is how could this CMake ever work in the first place when the linker wasn't explicitly set to the correct one tbh :thinking: It's so weird.

bugaevc:
> Without this flag the compiler uses the default linker it was configured to use.

But we're doing linking separately from compiling, Clang just produces `.o` files and then we have our own custom cross-linker from Apple's cctools

magiruuvelvet:
Running make in verbose mode (`make VERBOSE=1`) shows for me that CMake executes `clang` with all the `.o` files as input instead of Apple's linker. Than clang runs the wrong linker.

I think CMake uses the compiler to handle the linking instead of running the linker directly.

ahyattdev:
Could this PR be modified with a conditional so that this change is only done if `lsb_release -i` says Gentoo? Don't want to risk this breaking something on other distros that don't need it.

Also, since you got it working on Gentoo could you kindly provide a command to install dependencies that we can put in the build instructions list?

Examples for other distros are here: https://wiki.darlinghq.org/build_instructions

CuriousTommy:
@ahyattdev This patch actually fixed an issue I was having on Manjaro ARM (where the wrong linker was used).

ahyattdev:
@CuriousTommy OK I see. Upon further review I see why this is necessary.

PR #594, opened 2020-01-02
Fix DEB Build

### LINKED ISSUE #585: Download binary on Release
I failed trying to compile darling into my server at digital ocean. It ran out of disk space. 
Would it be possible to provide pre-built binaries, either verified or provided by a CI-CD service?

TheBrokenRail:
Fixes #585

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,5 @@ build
 rpm/SOURCES
 rpm/RPMS
 rpm/SRPMS
+
+debbuild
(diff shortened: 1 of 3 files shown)


ahyattdev:
Upon testing this the package build fails with this:

```
make[2]: *** No rule to make target 'printwrapgen'.  Stop.
```

Investigating now

ahyattdev:
That call to `make` was there before, so not the fault of this PR.

ahyattdev:
Looks like there is an open issue for this  #454

PR #612, opened 2020-01-05
Don't run scripts when building DEB

TheBrokenRail:


TheBrokenRail:
Any comments?

ahyattdev:
looks good to me

PR #623, opened 2020-01-10
Produce Upstream Tarball (DSC)

TheBrokenRail:


TheBrokenRail:
This is for the DSC build (source package) run before ```debuild``` but after ```debian/changelog``` is generated so the build tree should be clean.

ahyattdev:
OK, but those who have a previous `build` folder that are using this to get the Debian source package will just need to be aware that it will scoop that up too.

PR #624, opened 2020-01-10
Use .gitignore DSC

TheBrokenRail:


ahyattdev:
Very cool that tar can do that!

PR #627, opened 2020-01-11
--exclude-vcs Breaks the DSC Build

TheBrokenRail:
Because some submodues have files that would have been gitignored in the main repo, but aren't in their repo, the build fails, because the archive and folder are different.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -3,7 +3,7 @@ Maintainer: Lubos Dolezel <lubos@dolezel.info>
 Section: misc
 Priority: optional
 Standards-Version: 4.4.0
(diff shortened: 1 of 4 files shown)


TheBrokenRail:
Any updates?

ahyattdev:
How many submodules have their own .gitignore?

If it isn't too many a simple workaround could be taking the contents of each submodule .gitignore and adding them to the main .gitignore.

For example, the .gitignore of src/lkm would go into the main .gitignore as:

```
src/lkm/*.o
src/lkm/*.cmd

src/lkm/.tmp_versions
src/lkm/Module.symvers
src/lkm/*.ko
src/lkm/*.mod.c
src/lkm/modules.order
src/lkm/.cache.mk

src/lkm/*.o.ur-safe
```

TheBrokenRail:
The big issue is that some files that might be ignored in the main repo, exist in the sub-modules, so the tar archive, does not have some files, which the main directory has, which causes the build to fail.

PR #628, opened 2020-01-13
Fix Build-Depends

TheBrokenRail:
The ```python2``` package is only available on Ubuntu 19.04 and above and the ```python``` package is always available and also Python 2.


ahyattdev:
LGTM, we will need to change it back when distros start making `python` default to `python3`

TheBrokenRail:
For now both Ubuntu 19.10 and 18.04 have ```python``` as ```python2```.

Issue #425, opened 2018-09-05
Implement the open command


ahyattdev:

bugaevc:
See #43

LubosD:
CoreServices (LaunchServices) APIs used to determine what to do / take action:
* LSFindApplicationForInfo / LSCopyApplicationURLsForBundleIdentifier
* LSGetApplicationForInfo / LSCopyDefaultApplicationURLForContentType
* LSOpenURLsWithRole (uses Apple Events on macOS).

Issue #43, opened 2013-08-13
Implement LaunchServices


LubosD:
Implement [LaunchServices](https://developer.apple.com/library/mac/documentation/Carbon/Reference/LaunchServicesReference/Reference/reference.html) based on `xdg-mime` and `xdg-open`.

bugaevc:
Braindump:
* We need a way to run commands on the host (in this case, `xdg-open` and whatever it ends up starting)
    * Some "reverse shellspawn" process? Uh, I should get back to that "darling.cpp" project sometime
    * Another use-case: using iTerm2, asking it to run `darling-host-shell` (or whatever it gets called) instead of container's bash in new tabs/sessions
* We need the reverse, to be able to register apps "installed" (moved to `/Applications`?) under Darling with the Linux mime database (`mimeapps.list`)
    * Generate `.desktop` files from `Info.plist` files
       * E.g. `Exec=env DPREFIX=/tmp/newdarlingprefix darling run /Applications/AwesomeApp.app/Contents/MacOS/AwesomeApp %U`
       * Or should we only do this for the default prefix (`~/.darling`)?
    * Leave it to the DE to actually modify `mimeapps.list`
    * Have a launchd service that watches `/Applications` (using `WatchPaths` — does it even work under Darling?) and updates `.desktop` files
* Implement (or find a suitable implementation of) a simple file manager in Cocoa (Darling Finder)
    * Install it in `/Applications/Finder.app` (or `/System/Library/CoreServices/Finder.app`?)
    * Make it the default `.dmg` handler (`hdiutil attach` + open that location)
        * So people can unpack, install (drag to `/Applications`) and start using apps without ever manually running `darling shell`
    * Alternatively (or can we combine both of these somehow?), write a GNOME Software plugin ([tutorial](https://developer.gnome.org/gnome-software/stable/plugins.html), [examples](https://gitlab.gnome.org/GNOME/gnome-software/tree/master/plugins)) and KDE Discover backend ([examples](https://github.com/KDE/discover/tree/master/libdiscover/backends)); then opening a DMG would take the user directly into the familiar system software mamanger app where they can install/uninstall the app with just one button click
        * For this, it'd be nice to provide description, screenshots, version info, license and other metadata about the app (see [AppStream docs](https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html)). Some of this (icon, version, category) we can get from `Info.plist`, is there any source to get the rest?
* Same for `Installer.app`/`.pkg`

[...]

PR #661, opened 2020-02-16
Add open Command

TheBrokenRail:
This should make it easier for users to open GUI programs. Note, currently it is very basic, and does not support much functionality.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -285,6 +285,7 @@ add_subdirectory(external/libarchive/libarchive)
 add_subdirectory(external/apr)
 add_subdirectory(sandbox)
 add_subdirectory(xcselect)
+add_subdirectory(open)
 add_subdirectory(external/OpenDirectory)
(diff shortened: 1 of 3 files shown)


bugaevc:
I'm sorry but this is not how a proper `open` would look like. First of all it'd be a part of `LaunchServices`, not its own thing (so at the very least that directory should exist, and it should be in that directory).

Then `open` should open a lot more than just app-s. See https://github.com/darlinghq/darling/issues/43 and the existing code in https://github.com/darlinghq/darling/blob/master/src/frameworks/CoreServices/LaunchServices.cpp

TheBrokenRail:
Oh, OK.

mritunjaymusale:
I tried to run [this](https://reactstudio.com/) using `open` but I got 
`React%20Studio.app/Contents/MacOS/React%20Studio: Failed to spawn via launchd`

ahyattdev:
@mritunjaymusale unfortunately that's where it's stuck now. #425 tracks it

PR #667, opened 2020-02-16
Using Github Actions to build

stephengroat:
see https://github.com/stephengroat/darling/runs/449450231?check_suite_focus=true for example

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -79,7 +79,7 @@ jobs:
     - name: make deb src
       run: tools/makedeb --dsc
     - name: Move & compress DEBs
-      run: mkdir source && cp ../*~testing.* dist && tar -cJf source.tar.xz source
+      run: mkdir source && cp ../*~testing.* source && tar -cJf source.tar.xz source
     - name: Upload Artifacts
       uses: actions/upload-artifact@v1
       with:
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -26,14 +26,6 @@ jobs:
       run: |
         string(TIMESTAMP current_date "%Y-%m-%d-%H;%M;%S" UTC)
         message("::set-output name=timestamp::${current_date}")
-    - name: ccache info
-      run: ccache -s
-    - name: make deb src
-      run: tools/makedeb --dsc
-    - name: make deb
-      run: tools/makedeb
-    - name: debug
-      run: tree
     - name: ccache cache files
       uses: actions/cache@v1.1.0
       with:
@@ -41,3 +33,11 @@ jobs:
         key: ${ { matrix.config.name } }-ccache-${ { steps.ccache_cache_timestamp.outputs.timestamp } }
         restore-keys: |
           ${ { matrix.config.name } }-ccache-
+    - name: ccache info
+      run: ccache -s
+    - name: make deb src
+      run: tools/makedeb --dsc
+    - name: make deb
+      run: tools/makedeb
+    - name: debug
+      run: ls -lah ..
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -19,7 +19,7 @@ jobs:
         git submodule sync --recursive
         git -c "http.extraheader=$auth_header" -c protocol.version=2 submodule update --init --force --recursive --depth=1
     - name: install
-      run: sudo apt-get install cmake clang bison flex libfuse-dev libudev-dev pkg-config libc6-dev-i386 linux-headers-generic gcc-multilib libcairo2-dev libgl1-mesa-dev libtiff5-dev libfreetype6-dev git libelf-dev libxml2-dev libegl1-mesa-dev libfontconfig1-dev libbsd-dev libxrandr-dev libxcursor-dev ccache devscripts debhelper ninja-build
+      run: sudo apt-get install cmake clang bison flex libfuse-dev libudev-dev pkg-config libc6-dev-i386 linux-headers-generic gcc-multilib libcairo2-dev libgl1-mesa-dev libtiff5-dev libfreetype6-dev git libelf-dev libxml2-dev libegl1-mesa-dev libfontconfig1-dev libbsd-dev libxrandr-dev libxcursor-dev ccache devscripts debhelper ninja-build tree
     - name: Prepare ccache timestamp
       id: ccache_cache_timestamp
       shell: cmake -P {0}
@@ -32,10 +32,12 @@ jobs:
       run: tools/makedeb --dsc
     - name: make deb
       run: tools/makedeb
+    - name: debug
+      run: tree
     - name: ccache cache files
       uses: actions/cache@v1.1.0
       with:
-        path: .ccache
+        path: /home/runner/.ccache
         key: ${ { matrix.config.name } }-ccache-${ { steps.ccache_cache_timestamp.outputs.timestamp } }
         restore-keys: |
           ${ { matrix.config.name } }-ccache-
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -39,3 +39,8 @@ jobs:
       run: tools/makedeb --dsc
     - name: make deb
       run: tools/makedeb
+    - name: artifacts
+      uses: actions/upload-artifact@konradpabjan/v2-is-almost-here
+      with:
+        name: 'debs'
+        path: ../*.deb
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -79,9 +79,9 @@ jobs:
     - name: make deb src
       run: tools/makedeb --dsc
     - name: Move & compress DEBs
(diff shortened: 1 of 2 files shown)


TheBrokenRail:
Could you make this also build the DEB package? Also, ```.gitmodules``` used```..``` instead of the full path so you could use SSH or HTTPS and it would adjust automatically, for testing you should use ```git remote set-url``` for testing instead of changing ```.gitmodules```.

stephengroat:
@TheBrokenRail I can always try, let me try to address building the deb and using `git remote set-url` instead of editing `.gitmodules`

stephengroat:
don't see any instruction on how to use the `tools/makedeb` script, can you point me to any examples?

TheBrokenRail:
Just run ```tools/makedeb``` and DEBs are output in ```..``` (the odd directory seems to be a technical limitation).

ahyattdev:
I have hardware that could host a Jenkins instance for the project if we want faster builds.

.github/workflows/ccpp.yml
```diff
@@ -0,0 +1,32 @@
...
+    - name: install
+      run: sudo apt install cmake clang bison flex libfuse-dev libudev-dev pkg-config libc6-dev-i386 linux-headers-generic gcc-multilib libcairo2-dev libgl1-mesa-dev libtiff5-dev libfreetype6-dev git libelf-dev libxml2-dev libegl1-mesa-dev libfontconfig1-dev libbsd-dev libxrandr-dev libxcursor-dev
+    - name: mkdir
+      run:  mkdir build
+    - name: cmake
+      run: cd build && cmake ..
+    - name: make
+      run: cd build && make
+    - name: make lkm
+      run: make lkm
```
stephengroat: this doesn't seem to be a valid make target, am i running something wrong?

TheBrokenRail:
Can you upload the DEBs as artifacts? They are output in ```..```.

TheBrokenRail:
Here is some information about uploading artifacts: https://help.github.com/en/actions/configuring-and-managing-workflows/persisting-workflow-data-using-artifacts#uploading-build-and-test-artifacts.

stephengroat:
not going to be able to do a seperate `mkdir build && cd build && cmake .. && cmake` along with a `tools/makedeb`, disk space is limited.

to speed up the build, would using `ccache` on the system via `cmake` be an issue? something like this:

```
find_program(CCACHE_PROGRAM ccache)
if(CCACHE_PROGRAM)
    set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "${CCACHE_PROGRAM}")
endif()
```

TheBrokenRail:
Tris could also be configured to build a source package with ```tools/makedeb --dsc``` and then upload it to a Launchpad PPA using ```dput```.

TheBrokenRail:
You should generate the source package before the binary package so the source package doesn't contains the binary paclage data and has clean build tree. Also, do you have any plans to upload the DEB artifacts or upload the source package to Launchpad?

.github/workflows/ccpp.yml
```diff
@@ -0,0 +1,41 @@
+name: C/C++ CI
```
TheBrokenRail: Should the name be changed to something more specific like ```Darling CI```?

TheBrokenRail:
To upload to Launchpad you would need to upload a GPG key that can access it to the repository like this: https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets. The you need to make sure you set ```DEB_EMAIL``` to the email of that GPG key, and make sure you set ```DEB_DiSTRO``` to your target distro. Then, use ```debsign``` to sign the DSC like ```debsign ../darling_*_source.changes```. Finally, upload to Launchpad with ```dput -f ppa:<PPA> ../darling_*_source.changes```.

stephengroat:
@TheBrokenRail doing artifacts using Github Actions until the v2 is going to be hard since it doesn't support wildcard patterns yet (see https://github.com/actions/upload-artifact/tree/konradpabjan/v2-is-almost-here).

Want to wait to merge or just leave that feature out for right now?

TheBrokenRail:
Maybe copy the DEBs into a folder using ```cp``` and then make that folder the artifact?

TheBrokenRail:
Also, uploading to Launchpad wouldn't need Github Artifacts at all.

ahyattdev:
Merging this would add a dependency. What does `ccache` do for us here?

stephengroat:
> Merging this would add a dependency for the whole project. What does `ccache` do for us here?

@ahyattdev `ccache` is optional (only used if installed on the system, doesn't fail out otherwise) and provide c caching to speed up builds when possible

TheBrokenRail:
I meant like:
```yaml
- name: Move DEBs
  run: mkdir dist; cp ../*.deb dist
- name: Upload Artifacts
  uses: actions/upload-artifact@konradpabjan/v2-is-almost-here
    with:
      name: 'debs'
      path: dist
```

TheBrokenRail:
For Launchpad you could do:
```yaml
- name: Generate DSC
  run: tools/makedeb --dsc
  env:
    DEB_EMAIL: <Emaill for GPG Key>
    DEB_DISTRO: <Ubuntu Distro Launchpad Should Build For (for example: "eoan")>
- name: Import GPG Key
  run: echo ''${{ secrets.GPG_PRIVATE_KEY }}" | gpg --import > /dev/null
- name: Sign DSC
  run: debsign ../darling_*_source.changes
- name: Upload DSC
  run: dput -f ppa:<PPA> ../darling_*_source.changes
```

.github/workflows/ccpp.yml
```diff
@@ -0,0 +1,48 @@
...
+      run: tools/makedeb --dsc
+    - name: make deb
+      run: tools/makedeb
+    - name: cp debs
+      run: cp ../*.deb debbuild/
+    - name: artifacts
+      uses: actions/upload-artifact@konradpabjan/v2-is-almost-here
+      with:
+        name: 'debs'
+        path: debbuild/*.deb
```
TheBrokenRail: Why not just upload the whole directory itself instead of using wildcards like ```path: debbuild```, also where is ```debbuild``` created?

stephengroat: looks like just to 2 `deb`s by themselves are too large
https://github.com/stephengroat/darling/commit/c5b554465647f453f52be7f8b4c02cf8b1c6e51e/checks?check_suite_id=474725456

TheBrokenRail: Maybe try compressing them with:
```
- name: Move DEBs
  run: mkdir dist; cp ../*.deb dist
- name: Compress DEBs
  run: tar -cJf dist.tar.xz dist
- name: Upload Artifacts
  uses: actions/upload-artifact@v1
  with:
    name: 'debs'
    path: dist.tar.xz
```
Or just upload to Launchpad and not deal with Github storage limitations or do both!

bugaevc:
How about we start with convincing yourselves we actually want to use / rely on GitHub Actions / Launchpad?

TheBrokenRail:
The main argument for it would be that any Ubuntu/Debian users can just download the latest build of Darling instead of having to build it themselves. Also, it will automatically test if it can build correctly on Ubuntu.

tech234a:
Would uploading to Launchpad create a PPA for Ubuntu users to automatically update?

TheBrokenRail:
Yes

stephengroat:
Running out of disk space when trying to move and compress devs

Someone with PPA access will probably have to prototype the uploading with keys, maintainers can feel free to edit the branch and see the runs on https://github.com/stephengroat/darling/actions

LubosD:
If it is running out of disk space, we could use Travis instead. Travis provides 24GB, which should be enough for now.

stephengroat:
> If it is running out of disk space, we could use Travis instead. Travis provides 24GB, which should be enough for now.

Travis CI runtime limits are too low (1 hr) so the build will timeout before it completes

tech234a:
Maybe use something like Heroku: there, disk space is practically unlimited.

stephengroat:
artifacts are now uploading

https://github.com/stephengroat/darling/actions/runs/47920841

.github/workflows/ccpp.yml
stephengroat: changed

stephengroat:
let me know if there's anything else needed on this, trying to cleanup my PRs

stephengroat:
anything else needed for this? a squash or any specific comments i need to address?

ahyattdev:
@stephengroat I think this will work. We do have a Jenkins instance, but it's not "connected" to this repo. By that I mean having those green checkmarks next to the most recent commit of each push/PR.

Despite already having a Jenkins I think we should enable this too in order to verify that PRs build correctly or have a helpful green checkmark next to the most recent commit in master.

@LubosD @bugaevc any objections to merging this?

bugaevc:
> @LubosD @bugaevc any objections to merging this?

No technical objections, but we should decide whether CI-built DEBs is something we want to support (when people come complaining they don't work), and it would be nice to build for more than just the latest Ubuntu. IIRC, TeamCity at least had Debian.

PR #687, opened 2020-02-19
Fixing Debian build dependency with linux headers

r-lind:
Adding the closest equivalent to Ubuntu's linux-headers-generic in Debian, which is linux-headers-amd64.


ahyattdev:
OK I will merge and test

Issue #702, opened 2020-03-08
Error on Pop OS


JezWal:
I keep on getting this error code 

`-- Found dsymutil: /usr/bin/llvm-dsymutil-6.0
-- Compiler include path detected as /usr/lib/llvm-9/lib/clang/9.0.0/include/
CMake Warning (dev) at src/CMakeLists.txt:238 (add_subdirectory):
  The source directory

    /home/jez/darling/src/external/zlib

  does not contain a CMakeLists.txt file.

  CMake does not support this case but it used to work accidentally and is
  being allowed for compatibility.

  Policy CMP0014 is not set: Input directories must have CMakeLists.txt.  Run
  "cmake --help-policy CMP0014" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at src/CMakeLists.txt:304 (add_subdirectory):
  The source directory

    /home/jez/darling/src/external/vim

  does not contain a CMakeLists.txt file.

  CMake does not support this case but it used to work accidentally and is
  being allowed for compatibility.

  Policy CMP0014 is not set: Input directories must have CMakeLists.txt.  Run
  "cmake --help-policy CMP0014" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Error at src/CMakeLists.txt:307 (add_subdirectory):
  add_subdirectory given source "external/zip/zip/zip30" which is not an
  existing directory.

CMake Error at src/CMakeLists.txt:308 (add_subdirectory):
  add_subdirectory given source "external/zip/unzip/unzip-5.52" which is not
  an existing directory.

CMake Warning (dev) at src/CMakeLists.txt:316 (add_subdirectory):
  The source directory

    /home/jez/darling/src/external/top

  does not contain a CMakeLists.txt file.

  CMake does not support this case but it used to work accidentally and is
  being allowed for compatibility.

  Policy CMP0014 is not set: Input directories must have CMakeLists.txt.  Run
  "cmake --help-policy CMP0014" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at src/CMakeLists.txt:321 (add_subdirectory):
  The source directory

    /home/jez/darling/src/external/zsh

  does not contain a CMakeLists.txt file.

[...]

Issue #700, opened 2020-03-07
Uncomplete Configuration When Executing CMake Command


hassan91:
-- Configuring incomplete, errors occurred!
See also "/home/hassan/Desktop/libraries/darling/build/CMakeFiles/CMakeOutput.log".
--------------------------------------------------------------
CMakeOutput.log

The target system is: Linux -  - x86-64
The host system is: Linux - 4.15.0-88-generic - x86_64
Compiling the C compiler identification source file "CMakeCCompilerId.c" succeeded.
Compiler: /usr/bin/clang 
Build flags: 
Id flags:  

The output was:
0

Compilation of the C compiler identification source "CMakeCCompilerId.c" produced "a.out"

The C compiler identification is Clang, found in "/home/hassan/Desktop/libraries/darling/build/CMakeFiles/3.10.2/CompilerIdC/a.out"

Compiling the CXX compiler identification source file "CMakeCXXCompilerId.cpp" succeeded.
Compiler: /usr/bin/clang++ 
Build flags: 
Id flags:  

The output was:
0

Compilation of the CXX compiler identification source "CMakeCXXCompilerId.cpp" produced "a.out"

The CXX compiler identification is Clang, found in "/home/hassan/Desktop/libraries/darling/build/CMakeFiles/3.10.2/CompilerIdCXX/a.out"

Determining if the C compiler works passed with the following output:
Change Dir: /home/hassan/Desktop/libraries/darling/build/CMakeFiles/CMakeTmp

Run Build Command:"/usr/bin/make" "cmTC_3be68/fast"
/usr/bin/make -f CMakeFiles/cmTC_3be68.dir/build.make CMakeFiles/cmTC_3be68.dir/build
make[1]: Entering directory '/home/hassan/Desktop/libraries/darling/build/CMakeFiles/CMakeTmp'
Building C object CMakeFiles/cmTC_3be68.dir/testCCompiler.c.o
/usr/bin/clang    -o CMakeFiles/cmTC_3be68.dir/testCCompiler.c.o   -c /home/hassan/Desktop/libraries/darling/build/CMakeFiles/CMakeTmp/testCCompiler.c
Linking C executable cmTC_3be68
/usr/bin/cmake -E cmake_link_script CMakeFiles/cmTC_3be68.dir/link.txt --verbose=1
/usr/bin/clang      -rdynamic CMakeFiles/cmTC_3be68.dir/testCCompiler.c.o  -o cmTC_3be68 
[...]

PR #703, opened 2020-03-08
Update src/external/top to correct commit hash

facekapow:
This resolves an issue when cloning Darling that should help with #700 and #702.

The problem was that an non-existent commit hash was specified for `src/external/top`, leading to the following error when doing `git clone --recursive` or `git submodule update`:

```
fatal: remote error: upload-pack: not our ref 8bc7d87e088fdcd53c47ebf5460acc004f6b4fc4
fatal: the remote end hung up unexpectedly
Fetched in submodule path 'src/external/top', but it did not contain 8bc7d87e088fdcd53c47ebf5460acc004f6b4fc4. Direct fetching of that commit failed.
```

I'm not sure where this hash came from, but this commit updates the submodule to point to the latest commit in [darling-top](https://github.com/darlinghq/darling-top).


CuriousTommy:
> I'm not sure where this hash came from

https://github.com/darlinghq/darling/commit/71c5bf30efe757a0249e0f1825da63119478bb51 was likely the cause. @ahyattdev probably forgot to push his changes to the repo.

PR #714, opened 2020-03-22
Replace Current libffi Code With Apple's Version

CuriousTommy:
Just keep in mind that https://github.com/darlinghq/darling-ruby/pull/1 is needed as well. Otherwise, the build will fail.


bugaevc:
Can we have it as an external module?

CuriousTommy:
@bugaevc You're right. This would probably make more sense in an external module. If someone is able to open up one for me, I'll put the code there.

PR #719, opened 2020-03-27
Bump lkm module to 4040733

kelnos:
Fixes compilation on newer Linux kernels.

@LubosD looks like there are a couple other changes that this pulls in, but you wrote both of them and they seem safe...?


LubosD:
Yes, they are safe :-)

PR #723, opened 2020-03-30
Make AudioFileTools SHARED

nickooms:
Fixes CMake configuration fatal error on Deepin Linux


LubosD:
No, this is not supposed to be a shared library. The only proper way around would be to use TARGET_OBJECTS.

LubosD:
Problem has been solved differently.

PR #731, opened 2020-04-08
Stub `SCDynamicStoreKeyCreateComputerName`

facekapow:
Needed for `secd` and `trustd`.


bugaevc:
It's supposed to just copy the hostname (using the provided allocator), isn't it?

LubosD:
No, it's supposed to provide a key for notifications about hostname changes. So something that will not work without having a full configd :-/

I actually had the STUB already in place, but forgot to commit.

facekapow:
I would've implemented it the way `confid/SystemConfiguration.fproj/SCDHostName.c` implements it (all it needs is `SCDynamicStoreKeyCreate`, which is already stubbed), but I'm not clear on how the licensing works when copying functions from APSL code, so I decided to just stub it instead.

PR #737, opened 2020-04-13
Fix wrong path check in ImageLoaderMachO

disconnect3d:
This PR fixes a bug in `strncmp(installPath, "/usr/libexec/oah/Shims", 9)` - the size argument passed seems to be copied from a previous call and results in an incorrect path check as here only the `"/usr/libe"` prefix will be checked.

Please note that with my proposed change any directory path that starts with `"/usr/libexec/oah/Shims"` will be matched so `"/usr/libexec/oah/ShimsSOMETHING"` would also do.

I am not aware of this codebase, so maybe there should be a `"/"` at the end of this path?

### DIFF
--- a/src/dyld/src/ImageLoaderMachO.cpp
+++ b/src/dyld/src/ImageLoaderMachO.cpp
@@ -1388,7 +1388,7 @@ bool ImageLoaderMachO::needsAddedLibSystemDepency(unsigned int libCount, const m
 				// It is OK for OS dylibs (libSystem or libmath or Rosetta shims) to have no dependents
 				// but all other dylibs must depend on libSystem for initialization to initialize libSystem first
 				// <rdar://problem/6497528> rosetta circular dependency spew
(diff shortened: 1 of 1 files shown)


bugaevc:
Thank you for your contribution, but, no. This code comes from Apple; if they have fixed this bug in a newer version, we should update to that; if they haven't, we should leave the bug as is.

disconnect3d:
Good point. Reported further to https://github.com/opensource-apple/dyld/pull/4.

bugaevc:
I were much doubt Apple's going to read GitHub issues; file a Radar or whatever they call it these days if you really want this addressed.

ahyattdev:
I wonder how much effort updating our `dyld` would be

ahyattdev:
One consideration: did Apple mess with `i386` in dyld in 10.15 or is it more architecture-agnostic?

PR #760, opened 2020-04-20
Adding some basic functionality to sysctl_machdep

Balint66:


src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +77,132 @@ sysctl_handler(handle_brand_string)
...
+}
+sysctl_handler(handle_features)
+{
+    setup(1);
+
+    char *features[] = {"FPU","VME", "DE", "PSE", "TSC", "MSR", "PAE", "MCE", "CX8", "APIC", NULL,"SEP","PGE","MCA",
+                        "CMOV", "PAT", "PSE-36", "PSN", "CLFSH", "DS", "ACPI", "MMX", "FXSR", "SSE", "SSE2", "SS",
+                        "HTT", "TM", NULL, "PBE"};
+
+    char *out = (char*)(calloc(119, sizeof(char))); //"worst" case: we have all of it plus null termination
```
facekapow, line 178: We actually don't have `calloc` in `src/kernel` (it's an undefined symbol at link time); use `malloc` instead (along with `memset`, if you really need to zero the memory).
```suggestion
    char *out = (char*)(malloc(119 * sizeof(char))); //"worst" case: we have all of it plus null termination
    memset(out, 0, 119 * sizeof(char));
```
Don't forget to also add the declarations for these functions to the top of the file:
```c
extern void* malloc(__SIZE_TYPE__ len);
extern void* memset(void* s, int c, size_t n);
```

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +77,132 @@ sysctl_handler(handle_brand_string)
...
 
+sysctl_handler(handle_max_basic)
+{
+	setup(0);
+
+
+    __cpuid(level, eax, ebx, ecx, edx);
+
+
+	sprintf(old,"%d",eax);
```
facekapow, line 88: This (along with the other uses of `sprintf`) produces an "implicit function definition" error with Clang. Simple solution: declare it at the top of the file, like so:
```c
extern int sprintf(char* str, const char* format, ...);
```

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -28,7 +46,15 @@ const struct known_sysctl sysctls_machdep[] = {
...
 
-sysctl_handler(handle_brand_string)
+#ifndef setup
+#define setup(value)\
+            unsigned int level = 0;\
+            unsigned int eax = value;\
+            unsigned int ebx;\
+            unsigned int edx;\
+            unsigned int ecx
+
```
facekapow, line 56: Missing `#endif`

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +77,132 @@ sysctl_handler(handle_brand_string)
...
+}
+
+sysctl_handler(handle_brand_string)
+{
+    unsigned int level = 0;
+    unsigned int eax = 0x80000000;
+
+    unsigned int test[3];
+
+    __cpuid(level,eax,test, test+1,test+2);
```
facekapow, line 141: This function has a bunch of "invalid lvalue in asm output" errors when compiling, and this is because you can't use arrays as assembly arguments. What you can do is define `ebx`, `ecx`, and `edx` (like in the other functions) and use those as your arguments.
```suggestion
    unsigned int ebx = 0;
    unsigned int ecx = 0;
    unsigned int edx = 0;
    __cpuid(level, eax, ebx, ecx, edx);
```

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +77,132 @@ sysctl_handler(handle_brand_string)
...
+
+    eax = 0x80000001;
+
+    union
+    {
+        unsigned int brand[12];
+        char name[49];
+    } v;
+
+    __cpuid(level, eax, v.brand+0x1, v.brand+0x2, v.brand+0x3);
```
facekapow, line 154: Like I mentioned in my review above, all the `__cpuid` calls in this function error with "invalid lvalue in asm output". Taking into account my suggestion above, if you need to add the return values to the union, you can do:
```suggestion
    __cpuid(level, eax, ebx, ecx, edx);
    v.brand[0] = eax;
    v.brand[1] = ebx;
    v.brand[2] = ecx;
    v.brand[3] = edx;
```
Same goes for the other two uses of `__cpuid` below.

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +77,132 @@ sysctl_handler(handle_brand_string)
...
+    int size = sizeof(char);
+    int counter = 0;
+
+    for (int i = 0; i < 32; ++i)
+    {
+        if(edx>>i&1)
+        {
+            for(j=0;j< sizeof(features[i])/size; j++)
+            {
+                out[counter+j] = features[i][j]
```
facekapow, line 192: Missing semicolon:
```suggestion
                out[counter+j] = features[i][j];
```

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +77,132 @@ sysctl_handler(handle_brand_string)
...
+            counter = counter + j + 1;
+
+            out[counter] = ' ';
+
+            counter += 1;
+
+        }
+    }
+
+    copyout_string(out,(*char) old, oldlen);
```
facekapow, line 204: Typo:
```suggestion
    copyout_string(out,(char*) old, oldlen);
```

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +77,132 @@ sysctl_handler(handle_brand_string)
...
+
+    return 0;
+}
+sysctl_handler(handle_features)
+{
+    setup(1);
+
+    char *features[] = {"FPU","VME", "DE", "PSE", "TSC", "MSR", "PAE", "MCE", "CX8", "APIC", NULL,"SEP","PGE","MCA",
+                        "CMOV", "PAT", "PSE-36", "PSN", "CLFSH", "DS", "ACPI", "MMX", "FXSR", "SSE", "SSE2", "SS",
+                        "HTT", "TM", NULL, "PBE"};
```
facekapow, line 176: The second `NULL` is actually in the wrong place (it's at index 30, but it *should* be at index 20, according to [Wikipedia](https://en.wikipedia.org/wiki/CPUID#EAX=1:_Processor_Info_and_Feature_Bits)); plus, there's 31 elements, not 32.
```suggestion
                        "CMOV", "PAT", "PSE-36", "PSN", "CLFSH", NULL, "DS", "ACPI", "MMX", "FXSR", "SSE", "SSE2", "SS",
                        "HTT", "TM", "IA64", "PBE"};
```

src/kernel/emulation/linux/misc/sysctl_machdep.c
facekapow, line 178: With the suggested change to the table, `119` should be `124`.

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +77,132 @@ sysctl_handler(handle_brand_string)
...
+
+    int j = 0;
+    int size = sizeof(char);
+    int counter = 0;
+
+    for (int i = 0; i < 32; ++i)
+    {
+        if(edx>>i&1)
+        {
+            for(j=0;j< sizeof(features[i])/size; j++)
```
facekapow, line 190: `sizeof(features[i])` will always return `8`, since it's checking the size of the pointer, not the length of the string. Use this:
```suggestion
            for(j=0;j< strlen(features[i]); j++)
```
Also don't forget the declaration at the top of the file:
```c
extern __SIZE_TYPE__ strlen(const char* str);
```

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +77,132 @@ sysctl_handler(handle_brand_string)
...
+            {
+                out[counter+j] = features[i][j]
+            }
+
+            counter = counter + j + 1;
```
facekapow, line 195: You don't need to be skipping an extra byte:
```suggestion
            counter = counter + j;
```

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +77,132 @@ sysctl_handler(handle_brand_string)
...
+
+            out[counter] = ' ';
+
+            counter += 1;
+
+        }
+    }
+
```
facekapow, line 203: Null-terminate the string:
```suggestion
out[counter] = '\0';
```

src/kernel/emulation/linux/misc/sysctl_machdep.c
bugaevc: No, don't allocate memory on the heap at all. That's a weird thing to do for syscalls. Preallocate the memory, or allocate it on stack.

src/kernel/emulation/linux/misc/sysctl_machdep.c
bugaevc: No, you cannot use libc functions in syscalls. We have https://github.com/darlinghq/darling/blob/master/src/kernel/emulation/linux/simple.h for string formatting

src/kernel/emulation/linux/misc/sysctl_machdep.c
bugaevc: And it shouldn't be hardcoded in the first place

src/kernel/emulation/linux/misc/sysctl_machdep.c
bugaevc: Same here, don't call into libSystem, use `__simple_strlen`

src/kernel/emulation/linux/misc/sysctl_machdep.c
Balint66: Okay, but I'm not sure what do you mean by that @bugaevc it's my first time creating code for a kernel module.

src/kernel/emulation/linux/misc/sysctl_machdep.c
Balint66: done

src/kernel/emulation/linux/misc/sysctl_machdep.c
Balint66: done (then I went with for loops)

src/kernel/emulation/linux/misc/sysctl_machdep.c
bugaevc: This is not the kernel module, that would be https://github.com/darlinghq/darling-newlkm 😀

This is userspace syscall emulation

bugaevc: Basically I mean you cannot/shouldn't use `malloc/free`. Use `alloca`, or just static arrays.

src/kernel/emulation/linux/misc/sysctl_machdep.c
Balint66: I made it according to [this site](https://c9x.me/x86/html/file_module_x86_id_45.html)  and there the 30 is the reserved one. The 32 elements are good because the counting starts with 0 and ends with 31.

src/kernel/emulation/linux/misc/sysctl_machdep.c
Balint66: Okay and sorry. I thought it was the kernel module but after that, I realised that it's not because it's a separate repository. I'll use `alloca` then.

src/kernel/emulation/linux/misc/sysctl_machdep.c
Balint66: done

src/kernel/emulation/linux/misc/sysctl_machdep.c
Balint66: Oh, now I see. I had 31 elements, not 32.

src/kernel/emulation/linux/misc/sysctl_machdep.c
Balint66: Done

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +80,139 @@ sysctl_handler(handle_brand_string)
...
+    char *out = ""; //"worst" case: we have all of it plus null termination
+
+    __cpuid(0,eax,ebx,ecx,edx);
+
+    int j = 0;
+    int counter = 0;
+
+    for (int i = 0; i < 32; i++)
+    {
+        if(edx>>i&1)
```
facekapow, line 186: It turns out that the CPU can actually set the bits that are supposed to be reserved, so you need to explicitly skip these bits. e.g.:
```suggestion
        if (i == 10 || i == 2)
          continue;
        if(edx>>i&1)
```

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +80,139 @@ sysctl_handler(handle_brand_string)
...
+    v.name[48] = 0;
+    copyout_string(v.name, (char*) old, oldlen);
+
+    return 0;
+}
+sysctl_handler(handle_features)
+{
+    setup(1);
+
+    char *features[] = {"FPU","VME", "DE", "PSE", "TSC", "MSR", "PAE", "MCE", "CX8", "APIC", NULL,"SEP","MTRR","PGE",
```
bugaevc, line 173: Don't create `char *` (mutable pointers) to string literals.

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +80,139 @@ sysctl_handler(handle_brand_string)
...
+                        "MCA", "CMOV", "PSE-36", "PSN", "CLFSH", NULL, "DS", "ACPI", "MMX", "FXSR", "SSE", "SSE2", "SS",
+                        "HTT", "TM","IA64","PBE"};
+
+    char *out = ""; //"worst" case: we have all of it plus null termination
```
bugaevc, line 177: Ditto

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +80,139 @@ sysctl_handler(handle_brand_string)
...
+
+    int j = 0;
+    int counter = 0;
+
+    for (int i = 0; i < 32; i++)
+    {
+        if(edx>>i&1)
+        {
+						int len = __simple_strlen(features[i]);
+						int out_len = ____simple_strlen(out);
```
bugaevc, line 189: Wait, do both `__simple_strlen` and `____simple_strlen` exist?

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +80,139 @@ sysctl_handler(handle_brand_string)
...
+						char *new_out = alloca((len+out_len+1)*sizeof(char))
```
bugaevc, line 190: Don't `alloca` in a loop, that's a bad idea 'cause you're only going to free it when you exit the function.

`sizeof(char)` is always 1, it's not idiomatic to write it explicitly.

Why don't you just write the strings directly to the caller's buffer, instead of building the concatenated string on the stack and using `copyout` at the end?

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +80,139 @@ sysctl_handler(handle_brand_string)
...
 
+sysctl_handler(handle_max_basic)
+{
+	setup(0);
+
+
+    __cpuid(level, eax, ebx, ecx, edx);
+
+
+	sprintf(old,"%d",eax);
```
facekapow, line 91: This can cause actually cause segfaults because `sysctl` can be called with a `NULL` value for `old`, in order to determine the number of characters necessary. `copyout_string` already handles this, but doing `__simple_sprintf` (or `sprintf`) directly into the string can cause a segfault.

@bugaevc suggested that you could create a small stack-allocated buffer for the string, `__simple_sprintf` to that, and then use `copyout_string` on it. e.g.:
```suggestion
	char tmp[64]; // probably more than enough
	__simple_sprintf(tmp, "%d", eax);
	copyout_string(tmp, (char*)old, oldlen);
```

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +80,139 @@ sysctl_handler(handle_brand_string)
...
+    copyout_string(v.name, (char*) old, oldlen);
+
+    return 0;
+}
+sysctl_handler(handle_features)
+{
+    setup(1);
+
+    char *features[] = {"FPU","VME", "DE", "PSE", "TSC", "MSR", "PAE", "MCE", "CX8", "APIC", NULL,"SEP","MTRR","PGE",
+                        "MCA", "CMOV", "PSE-36", "PSN", "CLFSH", NULL, "DS", "ACPI", "MMX", "FXSR", "SSE", "SSE2", "SS",
```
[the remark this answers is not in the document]
Balint66: Now I just feel dumb. :sweat_smile:

src/kernel/emulation/linux/misc/sysctl_machdep.c
Balint66: typo. :sweat_smile:

src/kernel/emulation/linux/misc/sysctl_machdep.c
Balint66: Yeah, good point, actually I was thinking that "how do I know that the `old` argument has enough memory allocated", but that should have that much if it is expected. Should I call `copyout_string` on it with an empty string first or it is a ~~very~~ bad idea?

bugaevc: No, I wouldn't call `copyout_string` at all here. Just output these string pieces to `out` while there's enough space (if `out` is not null in the first place), and track the length yourself as you do now.

Balint66: Do you mean `old`? That is the actual output variable.

bugaevc: Yes, I mean `old` of course.

Balint66: Okay, I'll do that. Thank you!

Balint66:
Moral of the story: tried out three IDEs for Darling today, Atom is not bad, but Codium wins.

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +89,131 @@ sysctl_handler(handle_brand_string)
...
+sysctl_handler(handle_family)
+{
+    setup(1);
+
+    __cpuid(level, eax, ebx, ecx, edx);
+
+    eax = eax >> 7;
+    eax &= 15;
+
+    copyout_string(old,"%d",eax);
```
facekapow, line 116: I think you meant `copyout_int`, right?

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +80,139 @@ sysctl_handler(handle_brand_string)
...
+    sprintf(old,"%d",eax);
+
+    return 0;
+}
+
+sysctl_handler(handle_brand_string)
+{
+    setup(0x80000000);
+
+    __cpuid(level,eax,eabx, eacx, eadx);
```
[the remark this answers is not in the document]
facekapow: Still unresolved

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +89,131 @@ sysctl_handler(handle_brand_string)
...
+
+    v.name[48] = 0;
+    copyout_string(v.name, (char*) old, oldlen);
+
+    return 0;
+}
+sysctl_handler(handle_features)
+{
+
+    if(old != NULL)
```
facekapow, line 183: Like I mentioned before, `sysctl` can be called with a `NULL` value for `old` in order to determine the length of the output string. This is partially correct (it avoids segfaults due to `NULL` values for `old`), but you don't have a branch for actually determining and outputting the length of the output string.

You could do this by moving lines 185 through 191 outside this `if` and add an `else` to the condition that is executed when `old` is `NULL`. In that branch, you just have to add up the lengths of the features that are set (plus the spaces in between).

That would look something like this:
```c
 else {
        int len = 0;
        for (int i = 0; i < 32; i++)
        {

            if(i == 10 || i == 20)
                continue;
            if(edx>>i&1) {
                if (len != 0)
                    len++;
                len += __simple_strlen(features[i]);
            }
        }
        if (oldlen)
            *oldlen = len;
    }
```

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +89,131 @@ sysctl_handler(handle_brand_string)
...
+        for (int i = 0; i < 32; i++)
+        {
+
+            if(i == 10 || i == 20)
+                continue;
+
+            if(edx>>i&1 && current_old_len < oldlen)
+            {
+                int len = __simple_strlen(features[i]);
+                __simple_sprintf(old,"%s %s",old,features[i]);
```
facekapow, line 205: You're adding an extra space here when the string is still empty.
```suggestion
                if (current_old_len > 0)
                  __simple_sprintf(old,"%s %s",old,features[i]);
                else
                  __simple_sprintf(old,"%s",features[i]);
```

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +89,131 @@ sysctl_handler(handle_brand_string)
...
+        int counter = 0;
+        int current_old_len = __simple_strlen(old);
+
+        for (int i = 0; i < 32; i++)
+        {
+
+            if(i == 10 || i == 20)
+                continue;
+
+            if(edx>>i&1 && current_old_len < oldlen)
```
facekapow, line 202: You're comparing an integer with a pointer here; do:
```suggestion
            if(edx>>i&1 && current_old_len < *oldlen)
```

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -57,6 +57,15 @@ const struct known_sysctl sysctls_machdep[] = {
             unsigned int ecx
 #endif
 
+
+inline void copyout_int(int value, char* to_copy, int to_copy_leng)
```
bugaevc, line 61: `static`

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -28,7 +48,25 @@ const struct known_sysctl sysctls_machdep[] = {
...
+#define setup(value)\
+            unsigned int level = 0;\
+            unsigned int eax = value;\
+            unsigned int ebx;\
+            unsigned int edx;\
+            unsigned int ecx
+#endif
+
+
+inline void copyout_int(int value, char* to_copy, int to_copy_leng)
```
facekapow, line 61: `to_copy_leng` is actually supposed to be a pointer:
```suggestion
inline void copyout_int(int value, char* to_copy, int* to_copy_leng)
```

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +89,131 @@ sysctl_handler(handle_brand_string)
...
+    return 0;
+}
+sysctl_handler(handle_features)
+{
+
+    if(old != NULL)
+    {
+        setup(1);
+
+        char features[][] = {"FPU","VME", "DE", "PSE", "TSC", "MSR", "PAE", "MCE", "CX8", "APIC", "","SEP","MTRR","PGE",
```
facekapow, line 187: Can't have an array of arrays of undefined length; use:
```suggestion
        cosnt char* features[] = {"FPU","VME", "DE", "PSE", "TSC", "MSR", "PAE", "MCE", "CX8", "APIC", "","SEP","MTRR","PGE",
```

src/kernel/emulation/linux/misc/sysctl_machdep.c
bugaevc: `leng` is an unusual way to spell "length", it should be either `len` or `length`.

And it also shouldn't be an `int`, it should be `size_t` or `unsigned long`.

src/kernel/emulation/linux/misc/sysctl_machdep.c
bugaevc, line 205: No point in printing `old` over itself — just shift a pointer appropriately to append to the already printed part.

src/kernel/emulation/linux/misc/sysctl_machdep.c
bugaevc: You actually can, like this:
`char features[][5]`
or whatever the max size is. But I'm not sure that's better than `const char* features[]`

facekapow: Well yeah, I meant `char[][]` isn't valid because the values inside the parent array have to have the same type and length. That's perfectly fine (in this case, it would be `char[][7]` to include the null terminator).

But we don't need to copy constant strings into an array since we're not going modify them anyways.

bugaevc: `static const char features[][7]` will actually generate one contiguous character buffer and will not do any copying.

src/kernel/emulation/linux/misc/sysctl_machdep.c
Balint66: in the loop `old[counter+j] = features[i][j];` is sufficent or should I cast old to `char*`?

bugaevc: No, you still have to use `__simple_strncpy()` or something like that to copy the string.

Balint66: ~~As I see~~ We don't have `__simple_strcnp()` so the only thing left to use is still `__simple_sprintf()`.

bugaevc: You could add `__simple_strcnpy()` yourself, or actually, `strncpy()` is simple (no pun intended) enough you could just use the real one.

Balint66: Okay, added. Another thing: should I use `strncpy` to add the space between or just `old[counter] = ' ';`? I'd think that the first methode would be an overkill just to add a single character.

bugaevc: Yes, just assigning a single character is fine (but don't forget to keep track of the length).

src/kernel/emulation/linux/misc/sysctl_machdep.c
Balint66: Oh, I missed that one. Just a minute, next commit should fix it.

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +91,164 @@ sysctl_handler(handle_brand_string)
...
+                counter = counter + len;
+                
+                if(counter < *oldlen)
+                {
+
+                old[counter] = ' ';
+
+                counter++;
+
+                }
```
facekapow, line 220: This should be moved before the `strncpy`, and the check should be `counter > 0`. Reason: it's easier to do this check and conditionally add the space and increment the counter *before* calculating the offset of the new string (otherwise the `strncpy` would need a conditional offset for when to add a space at the beginning). Also, if you move this before the `strncpy`, then you don't need the `counter < *oldlen` check (since you already did that in the parent conditional); instead you can just check for when to insert the space (`counter > 0`).

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +91,164 @@ sysctl_handler(handle_brand_string)
...
+
+            }
+        }
```
facekapow, line 223: Null-terminate the string:
```suggestion
        }
        if (counter < *oldlen)
          outstr[counter] = '\0';
```

@bugaevc, should we also be assigning the generated string length to `*oldlen`? It looks like this is what `copyout_string` does.

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +91,164 @@ sysctl_handler(handle_brand_string)
...
+        {
+
+            if(i == 10 || i == 20)
+                continue;
+
+            if(edx>>i&1 && counter < *oldlen)
+            {
+                int len = __simple_strlen(features[i]);
+                
+                strncpy(old,features[i]);
```
facekapow, line 209: This is `strncpy`, not `strcpy` (`strncpy` is safer, though; good choice 👍), you need a maximum length argument.

Also, you need to offset the destination pointer by the correct amount. Note that this suggestion is assuming that you move the conditional below above this `strncpy`, like I suggested. Otherwise, you would need to add 1 to your offset if a space is going to be added.
```suggestion
                // the length argument here is essentially `min(len, *oldlen - counter)`
                strncpy(outstr + counter, features[i], len < (*oldlen - counter) ? len : (*oldlen - counter));
```
Don't forget to replace the declaration of `strcpy` with:
```c
extern char *strncpy(char *dest, const char *src, __SIZE_TYPE__ n);
```

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +91,164 @@ sysctl_handler(handle_brand_string)
...
+                            "HTT", "TM","IA64","PBE"};
+
+        __cpuid(0,eax,ebx,ecx,edx);
+
+    if(old != NULL)
+    {
+
+        int counter = 0;
+        int j = 0;
+
```
facekapow, line 198: `old` is a `void*`. Create a `char*` from it:
```suggestion
        char* outstr = old;
```

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +91,164 @@ sysctl_handler(handle_brand_string)
...
+            if(i == 10 || i == 20)
+                continue;
+
+            if(edx>>i&1 && counter < *oldlen)
+            {
+                int len = __simple_strlen(features[i]);
+                
+                strncpy(old,features[i]);
+
+                counter = counter + len;
```
facekapow, line 211: After moving the space conditional above the `strncpy`, this should be wrapped in the conditional that the space addition branch currently has (`counter < *oldlen`). e.g.:
```c
if (counter < *oldlen) {
  // modified `strncpy` call from suggestion above

  counter = counter + len;
}
```

src/kernel/emulation/linux/misc/sysctl_machdep.c
bugaevc: Could you also rename `counter` to something more descriptive?

src/kernel/emulation/linux/misc/sysctl_machdep.c
bugaevc: Yes, if it's not null.

src/kernel/emulation/linux/misc/sysctl_machdep.c
Balint66: I've renamed `counter` to `current_length` and added `if(current_length)` to the begining to add the space.

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -1,18 +1,40 @@
 #include "sysctl_machdep.h"
+#include "simple.h"
 #include <sys/errno.h>
+#include <alloca.h>
+
+extern char *strncpy(char *dest, const char *src);
```
facekapow, line 6: `strncpy`'s signature is `extern char *strncpy(char *dest, const char *src, __SIZE_TYPE__ n);`:
```suggestion
extern char *strncpy(char *dest, const char *src, __SIZE_TYPE__ n);
```

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +91,182 @@ sysctl_handler(handle_brand_string)
...
+
+    setup(1);
+
+        static const char features[][7] = {"FPU","VME", "DE", "PSE", "TSC", "MSR", "PAE", "MCE", "CX8", "APIC", "","SEP","MTRR","PGE",
+                            "MCA", "CMOV", "PAT", "PSE-36", "PSN", "CLFSH", "", "DS", "ACPI", "MMX", "FXSR", "SSE", "SSE2", "SS",
+                            "HTT", "TM","IA64","PBE"};
+
+        __cpuid(0,eax,ebx,ecx,edx);
+
+    if(old != NULL)
```
bugaevc, line 194: Can you please fix formatting? E.g. indentation is different between the few lines above; and there should be a space between `if` and `(` here, and there are more similar issues below. This all of course doesn't impact program execution, but it makes it harder to read & make sure it works as expected.

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +91,182 @@ sysctl_handler(handle_brand_string)
...
+    {
+
+        char *outsr = (char*)old;
+        int current_length = 0;
+        int j = 0;
```
bugaevc, line 199: What's this for?

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +91,182 @@ sysctl_handler(handle_brand_string)
...
+
+        for (int i = 0; i < 32; i++)
+        {
+
+            if(i == 10 || i == 20)
+                continue;
+
+            if(edx>>i&1 && current_length < *oldlen)
```
bugaevc, line 207: Can you please parenthesize this to make operation order explicit? `(edx >> i) & 1` or something like that. Better yet, if this is checking i'th bit, write it as `(edx & (1 << i))`.

You cannot dereference `oldlen` (or `outstr` as you do below) without null-checking them.

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +91,182 @@ sysctl_handler(handle_brand_string)
...
+
+                }
+                
+                if(current_length < *oldlen)
+                {
+
+                int len = __simple_strlen(features[i]);
+                
+                strncpy(outsr + current_length, features[i],
+                 (len < (*oldlen - current_length) ? len : (*oldlen - current_length)));
```
bugaevc, line 225: `strncpy()` won't copy more bytes then there is in the source string (so, more than `len`) anyway (though it will fill the remainder of the buffer with zeros), so you can just pass it the buffer size, here `*oldlen - current_length`.

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +91,182 @@ sysctl_handler(handle_brand_string)
...
+
+                current_length = current_length + len;
```
bugaevc, line 227: `+=`

src/kernel/emulation/linux/misc/sysctl_machdep.c
bugaevc: Ah, I see you already do the null check above.

src/kernel/emulation/linux/misc/sysctl_machdep.c
Balint66: @facekapow asked me to cast old to a char*. It seemed pointless to me but maybe it's for clarity.

bugaevc: No, I mean `j`

Balint66: Oh, forgot to delet it.

src/kernel/emulation/linux/misc/sysctl_machdep.c
Balint66: You mean `if (/*code*/)` or `if( /*code*/ )`?

bugaevc: `if (condition)`

Balint66: Okay. ~~I mean I'm the type of guy who uses spaces after the `(` and before `)`, that is why I'm asking then I'll write down to not to do that for Darling.~~

bugaevc: Not that Darling has one consistent code style...

Balint66: Still, if it's better for others to read it like that, I'll use it here. 😄

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -247,10 +243,10 @@ sysctl_handler(handle_features)
         for (int i = 0; i < 32; i++)
         {
 
-            if(i == 10 || i == 20)
+            if (i == 10 || i == 20)
                 continue;
 
-            if(edx>>i&1) 
+            if (edx>>i&1)
```
bugaevc, line 249: Change this check to `edx & (1<<i)` too

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +91,178 @@ sysctl_handler(handle_brand_string)
...
+        {
+
+            outsr[current_length] = '\0';
+
+        }
+
+    }
+    else
+    {
+        int len = 0;
```
bugaevc, line 241: Why use two different variables (`current_length` vs `len`)?

src/kernel/emulation/linux/misc/sysctl_machdep.c
```diff
@@ -51,3 +91,178 @@ sysctl_handler(handle_brand_string)
...
+            {
+
+                if (current_length)
+                {
+
+                    outsr[current_length] = ' ';
+
+                    current_length++;
+
+                }
```
bugaevc, line 213: Getting really nitpicky, but perhaps less blanks lines here (and around)?

Balint66: No blank lines, got before and after brackeys, got it.

PR #791, opened 2020-05-12
Changes needed for updated libplatform

CuriousTommy:
These are the following files that needed to be changed for https://github.com/darlinghq/darling-libplatform/pull/1. Before I merge in the change, I want to make sure that everyone is okay with these changes.


CuriousTommy:
Since **facekapow** is going to do a major code update, this PR will be pointless in the long run.

PR #813, opened 2020-05-26
Apple fonttools fixes

HinTak:
The last of the current  batch of fixes for Apple's font tools. Please review.

### DIFF
--- a/src/frameworks/ApplicationServices/ATS/src/ATS.c
+++ b/src/frameworks/ApplicationServices/ATS/src/ATS.c
@@ -22,6 +22,8 @@
 #include <stdio.h>
 #include <CoreFoundation/CoreFoundation.h>
 
+const CFStringRef kATSValidationActivationInfoKeep = CFSTR("ATSValidationActivationInfoKeep");
+
 static int verbose = 0;
 
 __attribute__((constructor))
--- a/src/frameworks/ApplicationServices/ATS/src/ATS.c
+++ b/src/frameworks/ApplicationServices/ATS/src/ATS.c
@@ -22,7 +22,38 @@
 #include <stdio.h>
 #include <CoreFoundation/CoreFoundation.h>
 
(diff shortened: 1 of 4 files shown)


CuriousTommy:
Normally, I would have the variables declared in a header file as well, but it seems like the other variables are not declared in a header file either...

HinTak:
I am mostly following neighboring code patterns. That said, I have xcode 11 around and they are in the xcode header.

CuriousTommy:
> they are in the xcode header

I guess we forgot to add it to the header. Since they are in the Xcode header, could you also define your changes in the header as well?

HinTak:
I'll put that in, as well as checking what values those strings are supposed to do too.

HinTak:
They are in two/three sub-framework headers, which are included by the main. I suppose if they are not part of apple opensource (I'll check), we are not supposed to copy verbatim, right?

CuriousTommy:
Yeah, don't copy non-opensource headers verbatim.

CuriousTommy:
@HinTak Is there anything else you need to do for this PR?

HinTak:
@CuriousTommy : in the code currently posted, just need to find the value of `kMDItemFinderComment` from genuine Metadata (Re-exported by CoreServices) . If you have access to genuine mac os x, I'd appreciate some help.
The other values from other frameworks I have checked myself.


Still on my private local repo is more strings like that from the Metadata framework, and also I haven't gotten round to add a new header for these as discussed above. But I see those as probably on-going and I may not be able to get to them any time soon.

So I'd say this is okay to merge except for the value of `kMDItemFinderComment`. I don't mind it either way - ie you merge it now if you can add/correct/confirm that single value - this would close quite a few issues I filed; or leave this open until I get to the further work. I am inclined to suggest the former, if you have that value handy, considering that it would close a few issues.

CuriousTommy:
I figured out the value, thank you for this PR!

HinTak:
Thanks!

PR #817, opened 2020-05-27
Fixing up `src/kernel` to support architectures that use the generic syscall header

CuriousTommy:
For newer architectures (like ARM64), they share the same [generic syscall header](https://github.com/torvalds/linux/blob/master/include/uapi/asm-generic/unistd.h). However, there has been a push by the Linux developers to reduce the number of duplicate syscalls in the newer generic syscall header. So Darling will need to work around this issue.

I decided to submit my lazy/bad first draft so that I can easily ask questions about my code.

- [x] ext/
	- [x] epoll_create.c | Setting the flag to 0 will cause `epoll_create1` to behave like `epoll_create`
	- [x] epoll_wait.c | Setting sigmask to NULL will cause `epoll_pwait` to behave like `epoll_wait`
	- [x] eventfd.c | Would setting 0 to the flags treat `eventfd2` the same as `eventfd`?
	- [x] file_handle.c | What is the difference between the syscall `open` and `openat`?
	- [x] inotify_init.c | Setting the flag to 0 will cause `inotify_init1` to behave like `inotify_init`
- [x] select
	- [x] poll.c | If sigmask is NULL, we only have to worry about adjusting the timeout argument.
	- [x] select.c | Setting sigmask as NULL will cause `pselect` to behave like `select`.
- [x] stat
	- [x] lstat.c
	- [x] mkfifo.c
- [x] time/utimes.c
- [x] unistd
	- [x] chmod_extended.c
	- [x] dup2.c | `man` page doesn't say anything about being able to set 0, but it seems to work fine in my personal testing.
	- [x] getpgrp.c
	- [x] mknod.c
	- [x] pipe.c | If the flag is set to 0, `pipe2` will behave the same as `pipe`.
- [x] vchroot_userspace.c
- [x] xattr/flistxattr.c

### DIFF
--- a/src/kernel/emulation/linux/ext/epoll_create.c
+++ b/src/kernel/emulation/linux/ext/epoll_create.c
@@ -8,15 +8,20 @@ extern long cerror(int __err);
 VISIBLE
 int epoll_create (int __size)
 {
-	int rv;
+	#if defined(__NR_epoll_create)
+		int rv;
 
-	rv = LINUX_SYSCALL(__NR_epoll_create, __size);
-	if (rv < 0)
-	{
-		cerror(errno_linux_to_bsd(-rv));
-		return -1;
-	}
+		rv = LINUX_SYSCALL(__NR_epoll_create, __size);
+		if (rv < 0)
+		{
+			cerror(errno_linux_to_bsd(-rv));
+			return -1;
+		}
 
-	return rv;
+		return rv;
+	#else
+		// The size argument in __NR_epoll_create is ignored
+		return epoll_create1(0);
+	#endif
 }
 
--- a/src/kernel/emulation/linux/ext/inotify_init.c
+++ b/src/kernel/emulation/linux/ext/inotify_init.c
@@ -8,15 +8,19 @@ extern long cerror(int __err);
 VISIBLE
 int inotify_init (void)
 {
-	int rv;
+	#if defined(__NR_inotify_init)
+		int rv;
 
-	rv = LINUX_SYSCALL(__NR_inotify_init);
-	if (rv < 0)
-	{
-		cerror(errno_linux_to_bsd(-rv));
-		return -1;
-	}
+		rv = LINUX_SYSCALL(__NR_inotify_init);
+		if (rv < 0)
+		{
+			cerror(errno_linux_to_bsd(-rv));
+			return -1;
+		}
 
-	return rv;
+		return rv;
+	#else
+		inotify_init1(0);
+	#endif
 }
 
--- a/src/kernel/emulation/linux/select/select.c
+++ b/src/kernel/emulation/linux/select/select.c
@@ -13,25 +13,29 @@ long sys_select(int nfds, void* rfds, void* wfds, void* efds, struct bsd_timeval
 
 long sys_select_nocancel(int nfds, void* rfds, void* wfds, void* efds, struct bsd_timeval* timeout)
 {
-	int ret;
-	struct linux_timeval ltv;
+	#if defined(__NR__newselect) || defined(__NR_select)
+		int ret;
+		struct linux_timeval ltv;
 
-	if (timeout != NULL)
-	{
-		ltv.tv_sec = timeout->tv_sec;
-		ltv.tv_usec = timeout->tv_usec;
-	}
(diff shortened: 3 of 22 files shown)


src/kernel/emulation/linux/ext/epoll_create.c
```diff
@@ -10,7 +10,11 @@ int epoll_create (int __size)
 {
 	int rv;
 
-	rv = LINUX_SYSCALL(__NR_epoll_create, __size);
+	#if __NR_epoll_create
+		rv = LINUX_SYSCALL(__NR_epoll_create, __size);
+	#else
+		rv = LINUX_SYSCALL(__NR_epoll_create1, 0);
```
CuriousTommy, line 16: Would it make more sense to call [epoll_create1](https://github.com/darlinghq/darling/blob/master/src/kernel/emulation/linux/ext/epoll_create1.c#L9) instead of writing another `LINUX_SYSCALL `?

src/kernel/emulation/linux/ext/inotify_init.c
```diff
@@ -10,7 +10,11 @@ int inotify_init (void)
 {
 	int rv;
 
-	rv = LINUX_SYSCALL(__NR_inotify_init);
+	#if __NR_inotify_init
+		rv = LINUX_SYSCALL(__NR_inotify_init);
+	#else
+		rv = LINUX_SYSCALL(__NR_inotify_init1, 0);
```
CuriousTommy, line 16: Same with this one, [inotify_init1](https://github.com/darlinghq/darling/blob/15c46505cddec16bdf856647f23f5e55db9dec59/src/kernel/emulation/linux/ext/inotify_init1.c#L9) exists.

src/kernel/emulation/linux/select/select.c
```diff
@@ -25,9 +25,12 @@ long sys_select_nocancel(int nfds, void* rfds, void* wfds, void* efds, struct bs
 #ifdef __NR__newselect
 	ret = LINUX_SYSCALL(__NR__newselect, nfds, rfds, wfds, efds,
 			(timeout != NULL) ? &ltv : NULL);
-#else
+#elif __NR_select
 	ret = LINUX_SYSCALL(__NR_select, nfds, rfds, wfds, efds,
 			(timeout != NULL) ? &ltv : NULL);
+#else
+	ret = LINUX_SYSCALL(__NR_pselect6, nfds, rfds, wfds, efds,
```
CuriousTommy, line 32: [sys_pselect](https://github.com/darlinghq/darling/blob/master/src/kernel/emulation/linux/select/pselect.c#L8) uses `pselect6`.

src/kernel/emulation/linux/ext/epoll_create.c
CuriousTommy: Eh... I decided to use it anyway

src/kernel/emulation/linux/xattr/flistxattr.c
```diff
@@ -31,7 +31,12 @@ long sys_flistxattr(int fd, char* namebuf, unsigned long size, int options)
 		char buf[64] = {0};
 		char path[4096] = {0};
 		__simple_sprintf(buf, "/proc/self/fd/%d", fd);
-		ret = LINUX_SYSCALL(__NR_readlink, buf, path, sizeof(path) - 1);
+		#if __NR_readlink
+			ret = LINUX_SYSCALL(__NR_readlink, buf, path, sizeof(path) - 1);
+		#else
+			// Path must be absolute
+			ret = LINUX_SYSCALL(__NR_readlinkat, -1, buf, path, sizeof(path) - 1);
```
CuriousTommy, line 38: @LubosD If the path is absolute, would it be fine if I hard code the `dirfd` argument to -1 (invalid value)? Most Linux syscalls that use the `dirfd` argument will ignore it if the path is absolute.

CuriousTommy: Nevermind, found out that `AT_FDCWD` exists.

CuriousTommy:
Okay, I believe everything should be in place. I probably merge this in a few days. Let me know if you need me to change anything.

PR #829, opened 2020-06-11
Auto-Calculate Darling Prefix in tools/uninstall

TheBrokenRail:
Also switched the shebang to ```/bin/sh```  since the script seemed ```sh``` compatible, and there are faster ```sh``` implementations than ```bash``` like ```dash```.

### DIFF
--- a/tools/uninstall
+++ b/tools/uninstall
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 set +e
 
 if [ "$(whoami)" != "root" ]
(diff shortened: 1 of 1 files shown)


HinTak:
Slightly confusing $PREFIX (where user files are) vs $DARLING_PREFIX (installation prefix). Suggest not using the latter.

TheBrokenRail:
```DARLING_PREFIX``` is just an internal variable. It cannot be changed by the outside world, it does not matter what it is called.

HinTak:
It is about readability...

TheBrokenRail:
I updated the variable name.

TheBrokenRail:
Any updates?

TheBrokenRail:
Any updates on this?

ahyattdev:
LGTM

PR #832, opened 2020-06-12
Add Issue Templates

TheBrokenRail:
This should make it easier for new people.


facekapow:
I'm going to go ahead and merge this, since everything look good and it should help new users (well, everyone, actually) create better issues with more useful information from the start. Thanks, @TheBrokenRail!

Issue #848, opened 2020-07-22
Darling wiped my external drive (twice...)


HinTak:
I don't know how it happened - but for some reason, I found that darling has wiped my external drive (under `/run/media/<myid>/diskname` - mounted by udisk2) twice, on `launchctl shutdown`.

The first time, about 6 weeks ago, was quite painful - 350GB from a 1TB USB drive formatted to ext4 was gone - and still suffering from it . Surviving files are either owned by root (not my user account) or in directories which I happened to not have write access (ie mode 444). 

The 2nd time, just now, is only a 32 GB drive about 2/3 full... All the files were mine, so all of it gone.

The first time it happened it was late at night, I was doing too many things on too many terminals, I thought I might have accidentally cut-paste some combination of `rm -rf ` (and I have been kicking myself for a month) - the 2nd time I am not doing anything else, other than just using da rling for about 39 minutes and quit.

What exactly does darling shell does to clean up the working directory overlay and private mounts???

This bug is too scary :-( .

bugaevc:
This is why we always tell you to never run Darling on the host, always in a VM. (Not this in particular, but the possibility of bugs like this.)

> What exactly does darling shell does to clean up the working directory overlay and private mounts???

Nothing. When the container exits, Linux cleans up the mounts automatically. You'll have to look into what launchd does, if this happens when you tell launchd to shut down.

HinTak:
I am guessing some bad interaction between darling and udisks2 mounted disks.

Oh, I was also using DPREFIX this time, and was probably last time too (last time I had more than 1 darling sessions, one in default `~/.darling`, and the other in `~/d1`). This time I was continuing with `~/d1` . But I had used darling briefly between the two events, don't remember which one I was using.

HinTak:
My "/" and "/home" are both LVM volumes with ext4 .

Obviously `/home` is visible as `/SystemRoot/home` (and `/run/media/myid/diskname` similar) - and at least I am glad it did not touch `/home`. 

Hmm, `udisks2` has a message about cleaning up `/run/media/myid/diskname` , which usually just means removing the empty directory mount point, which it previously creates , when disks are unmounted.

This bug is way too scary to have seen once, let alone twice...

HinTak:
[...]

PR #852, opened 2020-07-23
limiting recursive file-wiping to known ones created by darling

HinTak:
See #848

### DIFF
--- a/src/startup/darling.c
+++ b/src/startup/darling.c
@@ -932,7 +932,7 @@ static void wipeDir(const char* dirpath)
 
 		if (ent->d_type == DT_DIR)
 		{
-			wipeDir(path);
+			// wipeDir(path); /* Do not recurse! */
 			rmdir(path);
 		}
 		else
@@ -946,8 +946,14 @@ void darlingPreInit(void)
 {
 	// TODO: Run /usr/libexec/makewhatis
 	const char* dirs[] = {
-		"/var/tmp",
-		"/var/run"
+		"/var/tmp/launchd/sock",
+		"/var/tmp/launchd", /* directory to be removed, after their contents */
+		"/var/run/.systemStarterRunning",
+		"/var/run/memberd.pid",
+		"/var/run/shellspawn.sock",
+		"/var/run/syslog.pid",
+		"/var/run/syslog",
+		"/var/run/utmpx"
 	};
 
 	char fullpath[4096];


bugaevc:
I'm sorry, but no, we do actually want to wipe `/var/tmp` and `/var/run`, not just those files that launchd/shellspawn create when you start the container — anything any programs you might run would create there.

To fix #848, we need to figure out what has actually gone wrong. If it is indeed these lines that are causing your host files to be wiped, then why.

HinTak:
"/var/run" is on the bottom of the overlay so it does not need to go and likely won't go. I haven't figured out what creates "/var/tmp" itself, but I'd prefer not to touch it if unsure.

HinTak:
I'd beg to differ - it is simply not safe to do recursive delete, regardless of whether that causes #848 .

bugaevc:
I don't think I understand your point — how or why is it not safe to delete recursively?

HinTak:
Also darling's main executable is suid root... I'll really ask you, why should silent and automated recursive delete ever be done? And silently and without asking the user to confirm too?

bugaevc:
Because these are internal temp files that are normally cleaned up on every reboot, so that is what we do / emulate. You're not surprised your host's `/tmp` and `/run` start empty on each reboot, without asking for confirmation, are you? We do the same for container's tmp and run.

HinTak:
And the darling start-up code is a host process - any mistakes, it recurses down host paths...

bugaevc:
FWIW, we do want to not have to run it as setuid root, eventually. But we do check that you own the prefix before doing anything with it, including wiping those directories. If there's a bug there somewhere, we should get *that* fixed.

HinTak:
Well, unfortunately, as I wrote in the bug report, checking ownership does not help - the external drive was mounted by udisks2 on behalf of the console user, and so is the tmpfs used by gnome/dbus ...

HinTak:
I also see in the start up code you try to load the kernel module, which requires root privilege anyway.

bugaevc:
Well, we have no evidence that this wiping logic is actually to blame for clearing your drives; it's only a suspicion of mine. And I don't see any apparent bugs with it that would cause it to ignore `DPREFIX` and operate on the host directory structure.

A step forward would be for you to add some tracing to this code (at least print which directories it's trying to wipe) and try to repro the bug.

HinTak:
Actually I explicitly disabled systemd's unit to periodically wipe /tmp . systemd's defaults to remove any files in /tmp that has not been used within 10 days - I found that crazy.

HinTak:
To be honest I'd prefer not to try to "reproduce" this bug - you think seeing it twice and losing 350 GB in one, and another 20 GB in another, is not enough???

HinTak:
I'd put in every precaution, and every protection I can think of, like this pull, before I even use darling at all again!

bugaevc:
0th rule of Darling development: only ever use Darling on a VM, never on the host (or worse, on your daily driver system).

HinTak:
You should put that in the top level readme - see how many contributors you'd lose.

HinTak:
If I want to run a VM, I'd just run Mac OS X in kvm.

bugaevc:
It's not like we have tons of contributors to lose some... But it is [in the docs](https://docs.darlinghq.org/build-instructions.html): "Be prepared for kernel hangups and crashes, and run Darling on a virtual machine if possible." (And it used to be worded stronger, if memory serves me right.)

Look, I understand you're upset about losing your data — and I can empathize, I've once lost a whole drive too (by accidentally dd'ing to a wrong destination). If you don't want anything like that to happen to you again, don't run software you can't trust to behave without sandboxing it.

If you want to help us track down and fix this bug (if it even is our bug, and not something else on your system), the way to go about that would be to reproduce it (on a VM, of course) and see if it's darling.c wiping wrong directories, or launchd shutting down, or what.

HinTak:
It already happened twice. I'd prefer to see some precautionary code like this pull (not necessarily as is - you are free to improve on it, instead of just closing) to be included, before I even use darling again.

HinTak:
And it is not like that bit in recursive wiping code is carved in stone - even as recent as 6 months ago it was wiping things it should not https://github.com/darlinghq/darling/commit/ff12c564febb12cf9a744cf48c1e35b416b6d836#diff-cd6f834c4e69dc3152d667c814eb15c8

@LubosD so I'd really like this pull to be re-opened, and improved upon if needed, before any further usage.

PR #864, opened 2020-08-19
Update to work on Fedora 31 with newest version

andyneff:
Note, I didn't make Fedora 32 the default version, only because I don't currently have Fedora 32 to test the RPMs on. But I fully expect them to work. if someone uses them on Fedora 32


bugaevc:
Why change dnf to yum?

HinTak (commented):
`dnf` to `yum` is a regression - `dnf` is the future, and yum (dnf's emulation of yum) will be gone at some point. In fact I think yum does not work / emit a warning and abort on fedora 32, I think.

andyneff:
I was going to play with the idea of making it work on CentOS 7, which uses yum. Fedora had always accepted `yum` thus far...

However I was reading the docs and I see that kernel 5.0 is a now a minimum requirement, so I will abort that en devour and will restore it to dnf. Thanks for the feedback!

andyneff:
@bugaevc Yes, the _default_ OS is indeed Fedora. But there is an argument that can replace the OS, at build time. And I _was_ looking into the idea of supporting CentOS as an OS value too. It is always best to build the package for the exact OS you are running.

For example, Ubuntu 16.04 is based off of debian 9, but I've had cases where using a debian 9 deb failed (and debian 8 just happened to work barely). Point is matching the OS exactly makes sure it will work. where as using the LIKE OS is risky

But, this is all moot, as even CentOS 8 is based off of kernel 4.18, which is too old for a 5 requirement.

---

So, in the end, if you are running Fedora 31, all you need to do is run `docker-compose build; docker-compose run rpm`
If you are Fedora 32, then run `RPM_OS=fedora:32 docker-compose build; docker-compose run rpm`

And if you are running CentOS/RHEL, you don't meet minimum requirements, so it does not matter.

HinTak:
Just to correct - I believe earlier yum emits a warning; but yum on fedora 32 is a reimplementation based on dnf. Anyway, the whole point of docker is that you can run/build things in an environment which looks like the real thing, so you should build in a centos docker for centos, ubuntu docker for ubuntu, etc.

That said, yes, if centos is still on kernel 4.x, this is irrelevant, since darling does not run on 4.x kernels (it may have in the past , when 4.x was current) and there is little value in trying to support old kernels.

andyneff:
Due to the order of event, updating a kernel module (using dkms) would update the new kernel module, just to then remove it. This was an annoying bug I wasn't entirely sure how to fix.

This occurred due to the order yum (or rpm?) updates packages. First it installs the new package, then after that, it removes the old package.

New version
1. Pre-install script
2. Installs files
3. Post-install script

Old version

4. Pre-uninstall script
5. Uninstalls files
6. Post-uninstall script

So if the version numbers were the same, the post-uninstall (6) step would remove the kernel module that was added in step 3. Not ideal. 

I tried following the steps in a similar dkms issue (25)

andyneff:
All my tests proved successful.

@HinTak @bugaevc Are there any other changes you need me to make?

HinTak:
Looks alright. About the dkms  issue - I think there is a dkms service which runs on booting a new kernel, so a post-install step may not be needed? Also I have an impression that some dkms installation just leaves modules behind - it means eventually /lib/modules accumulates old unused modules.

PR #921, opened 2021-02-03
Make .gitmodules formatting consistent

theslimshaney:
Make .gitmodules formatting consistent


ahyattdev:
I tested this, did not break an existing cloned copy.

Issue #856, opened 2020-08-03
TARGET_OS_DRIVERKIT missing under Clang 12


rdrpenguin04:
I should mention that this is a git build of Clang and of CMake; let me know if you want more precise version information.

**Build Log**
Relevant section (omitted successful targets):
```
[13/15893] Building CXX object src/dyld/CMakeFiles/system_loader.dir/src/dyld2.o
FAILED: src/dyld/CMakeFiles/system_loader.dir/src/dyld2.o 
[...]

Issue #964, opened 2021-04-30
error: prefix attribute must be followed by an interface or protocol


jcgonzalezr92:
Hello All, 

This happened when i tried the make, I installed all dependencies and all was going ok until here:  

[ 12%] Building CXX object src/external/objc4/runtime/CMakeFiles/objc_obj.dir/maptable.mm.o
[ 12%] Building CXX object src/external/objc4/runtime/CMakeFiles/objc_obj.dir/hashtable2.mm.o
[ 12%] Building CXX object src/external/objc4/runtime/CMakeFiles/objc_obj.dir/NSObject.mm.o
/home/user/darling/src/external/objc4/runtime/NSObject.mm:2000:2: error: prefix attribute must be followed by an interface or protocol
@implementation _NSObject
 ^
1 error generated.
make[2]: *** [src/external/objc4/runtime/CMakeFiles/objc_obj.dir/build.make:89: src/external/objc4/runtime/CMakeFiles/objc_obj.dir/NSObject.mm.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:11542: src/external/objc4/runtime/CMakeFiles/objc_obj.dir/all] Error 2
make: *** [Makefile:130: all] Error 2

I was wondering if you have any advise or if you can see that i may have done something wrong please let me know.

Thanks and regards!

facekapow:
What commit are you trying to build and what version of clang are you using?

ivanpmartell:
I'm not OP, but I have the same error with the following:
commit 35534fcaa312b184145c95a543cf86c8b08d4e70

clang version 6.0.0-1ubuntu2 (tags/RELEASE_600/final)
Target: x86_64-pc-linux-gnu
Thread model: posix

Running on ubuntu bionic:
Distributor ID: Ubuntu
Description:    Ubuntu 18.04.5 LTS
Release:        18.04

HinTak:
That clang is way too old - current is 11 or 12? It is certainly 10+ . You'll have problem with old kernel version with ubuntu 18.04 too! Consider upgrading to 20.04 at least.

facekapow:
Yeah, certainly too old. [The docs say](https://docs.darlinghq.org/build-instructions.html#dependencies) Clang 9+ is required (latest Clang release is 12).

Issue #939, opened 2021-03-01
NSObject.o build error: prefix attribute must be followed by an interface or protocol


paianoa:
I'm trying to compile on a 64bit virtual machine with Debian10 and kernel 5.10 (4GB ram):

```
root@phptest:/home/andrea/darling/build# uname -a
Linux phptest 5.10.0-0.bpo.3-amd64 #1 SMP Debian 5.10.13-1~bpo10+1 (2021-02-11) x86_64 GNU/Linux

```
Error got:
```

[ 13%] Built target system_duct_firstpass
[ 13%] Building CXX object src/external/objc4/runtime/CMakeFiles/objc_obj.dir/NSObject.o
/home/andrea/darling/src/external/objc4/runtime/NSObject.mm:2000:2: error: prefix attribute must be followed by an interface or protocol
@implementation NSObject
 ^
1 error generated.
make[2]: *** [src/external/objc4/runtime/CMakeFiles/objc_obj.dir/build.make:89: src/external/objc4/runtime/CMakeFiles/objc_obj.dir/NSObject.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:11542: src/external/objc4/runtime/CMakeFiles/objc_obj.dir/all] Error 2
make: *** [Makefile:130: all] Error 2
```

facekapow:
What Clang version are you compiling with?

paianoa:
Clang is 6.0.1 as you can see below:
```
root@phptest:/home/andrea/darling/build# cmake ..
-- The C compiler identification is Clang 6.0.1
-- The CXX compiler identification is Clang 6.0.1
-- Check for working C compiler: /usr/bin/clang-6.0
-- Check for working C compiler: /usr/bin/clang-6.0 -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/clang++-6.0
-- Check for working CXX compiler: /usr/bin/clang++-6.0 -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- The ASM compiler identification is Clang
-- Found assembler: /usr/bin/clang-6.0
-- The ASM-ATT compiler identification is GNU
-- Found assembler: /usr/bin/as
-- Found dsymutil: /usr/bin/llvm-dsymutil-6.0
[...]

PR #980, opened 2021-06-24
Update CMake supported clang versions

JJTech0130:
+ ~~Darling doesn't build with clang versions less than 9, so there is no need to support them: https://github.com/darlinghq/darling/issues/939 and https://github.com/darlinghq/darling/issues/964.~~ **EDIT:** Fixed by https://github.com/darlinghq/darling/commit/4fbf800df4dfa6e6159b2f97c8a053c9489e9dce.
+ Clang 11 is easier to get than 9 or 10 (buster-backports only has 8 and 11) so we should add it for ease of use.
+ Clang 12 support seems to be resolved: https://github.com/darlinghq/darling/issues/856.

~~I also updated the docs to reflect this change: https://github.com/darlinghq/darling-docs/pull/25~~

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,8 @@ if (NOT DEFINED CMAKE_C_COMPILER)
 		"clang-7"
 		"clang-9"
 		"clang-10"
+		"clang-11"
+		"clang-12"
 )
 endif(NOT DEFINED CMAKE_C_COMPILER)
 
@@ -21,6 +23,8 @@ if (NOT DEFINED CMAKE_CXX_COMPILER)
 		"clang++-7"
 		"clang++-9"
 		"clang++-10"
+		"clang++-11"
+		"clang++-12"
 )
 endif(NOT DEFINED CMAKE_CXX_COMPILER)
 


bugaevc:
To be honest, I'd like to keep old versions of Clang working too. If something we wrote doesn't work with them, I'd consider it a bug. However, if it's something Apple wrote (like objc4), then there's not much we could do.

JJTech0130:
I mentioned some issues in https://github.com/darlinghq/darling-docs/pull/25, and they were closed with "your clang is too old".

bugaevc:
So `objc_nonlazy_class` looks like it can be trivially made to work on older Clang versions:

```objc

#if /* have new enough Clang */
__attribute__((objc_nonlazy_class))
#endif
@implementation Foo

#if /* have old Clang */
+ (void) load { }
#endif
```

(Empty load methods for `NSObject` and others is also what the older versions of objc4 use.)

cc @facekapow

facekapow:
Yeah, that look about right (not sure if that's all that older clangs would need, though). That's actually what `OS_OBJECT_NONLAZY_CLASS` and `OS_OBJECT_NONLAZY_CLASS_LOAD` in libdispatch do. The objc runtime would be the only one that needs to be updated (everything else uses libdispatch's macros).

JJTech0130:
Well, um, I think that's ok. We should still add to new versions. Can I remove commits from the pull request, or is that not possible?

facekapow:
Yes; you can reset your branch back to master, and then make new commits, and then force push to the branch.

JJTech0130:
Well, can't do that on mobile, and my pc's a bit stressed out trying to build darling... Will do it as soon as I finish.

facekapow:
Currently testing that potential fix for older clangs. I'm on Arch, though, so I can't test Clang 6 specifically (well, I could, but the AUR only has Clang 5 and then 7), only Clang 7.

JJTech0130:
OK. I have fixed the commits.

JJTech0130:
How should the docs be updated? The Debian 10 requirements specified `clang-6.0`, whilst there is a notice stating only clang 9+ is supported. Should the notice be changed? Removed? And should we leave the Debian 10 instructions or update them?

facekapow:
The docs should be updated to indicate that the minimum required Clang version is actually 6 now, not 9.

JJTech0130:
https://github.com/darlinghq/darling-docs/pull/26

PR #1062, opened 2021-11-15
fix: Various kernel fixes

trungnt2910:
Various fixes are made in order to ~~run `dotnet` properly~~ make `dotnet` half-runnable.

## proc_info:
- Added 3 additional flavors.
- Added more info for `_proc_pidinfo_tbsdinfo`

## waitid
- Marshalled the `siginfo_t` between Linux and BSD, allowing processes to access valid data.

## SIGCHLD
- Some processes check for explicitly ignored signals and avoid handling them (e.g. `dotnet`). As `SIGCHLD` is not configured with `SIG_IGN` on real MacOS, we should not do that on Darling, either.


LubosD:
LGTM. Anyone else wants to review?

facekapow (approved):
LGTM. My only concern was the `SIG_IGN` change affecting LLDB, but doing some testing with an old LLDB version (one that was actually working) shows that it still works fine.

trungnt2910:
@facekapow @LubosD Can you review the last commit? It's simply some reexported symbols in `CoreServices` to get some `dlsym`s working.

src/frameworks/CoreServices/reexport.exp
```diff
@@ -0,0 +1,33 @@
...
+_log2$fenv_access_off
+_modff
+_modff$fenv_access_off
+_nan
+_nanf
+_nextafterd
+_nextafterf
+_remquo
+_remquo$fenv_access_off
+dyld_stub_binder
```
facekapow, line 33: `dyld_stub_binder` isn't supposed to be re-exported; it's automatically added by `ld` for each library.
```suggestion
```

facekapow (changes requested):
Just a small issue with re-exporting `dyld_stub_binder`. Other than that, LGTM.

src/frameworks/CoreServices/reexport.exp
trungnt2910: Done. Forced pushed the change.

trungnt2910: It's a bit of an oversight when copy-pasting stuff.

facekapow (approved):

trungnt2910:
Thanks so much! This pull request wouldn't have been possible without your mentoring!

PR #1150, opened 2022-04-05
feat: Manage executable path in darlingserver

trungnt2910:
With darlingserver managing executable paths, `_proc_pidinfo_pathinfo` can get the path from darlingserver instead of procfs (which points to mldr instead of the real path).

Also, `mldr` now announces the executable path after loading and before starting the thread.

### DIFF
--- a/src/kernel/emulation/linux/misc/proc_info.c
+++ b/src/kernel/emulation/linux/misc/proc_info.c
@@ -646,11 +646,19 @@ static bool parse_smaps_firstline(
 static long _proc_pidinfo_pathinfo(int32_t pid, void* buffer, int32_t bufsize)
 {
 	struct vchroot_unexpand_args args;
+	uint64_t fullLength;
 
-	int rv = dserver_rpc_get_executable_path(pid, args.path, sizeof(args.path));
+	int rv = dserver_rpc_get_executable_path(pid, args.path, sizeof(args.path), &fullLength);
 
-	if (rv < 0)
-		return rv;
+	if (rv < 0) 
+	{
+		__simple_printf("dserver_rpc_get_executable_path failed internally: %d\n", rv);
+		__simple_abort();
+	}
+	else if (rv > 0) 
+	{
+		return -rv;
+	}
 
 	rv = vchroot_unexpand(&args);
 	if (rv != 0)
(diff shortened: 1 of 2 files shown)


src/startup/mldr/mldr.c
```diff
@@ -188,6 +188,8 @@ int main(int argc, char** argv, char** envp)
 		exit(1);
 	}
 
+	dserver_rpc_set_executable_path(filename, strlen(filename));
```
facekapow, line 191: Error handling, just in case (which is just exiting since we can't actually do anything without the server):
```suggestion
	if (dserver_rpc_set_executable_path(filename, strlen(filename)) < 0) {
		fprintf(stderr, "Failed to tell darlingserver about our executable path\n");
		exit(1);
	}
```

src/kernel/emulation/linux/misc/proc_info.c
```diff
@@ -645,20 +645,13 @@ static bool parse_smaps_firstline(
...
 	struct vchroot_unexpand_args args;
 
-	__simple_sprintf(path, "/proc/%d/exe", pid);
-
-	memset(buffer, 0, bufsize);
-	int rv = sys_readlink(path, buffer, bufsize - 1);
+	int rv = dserver_rpc_get_executable_path(pid, args.path, sizeof(args.path));
 
 	if (rv < 0)
 		return rv;
```
facekapow, line 653: Like I said [here](https://github.com/darlinghq/darlingserver/pull/1/files#r842329980), negative error codes from RPC calls indicate internal errors; we should abort on internal errors (maybe handling this should be moved into the RPC wrapper code later). However, positive error codes are okay (e.g. in this case, it'd okay to get positive `ESRCH`).
`_proc_pidinfo_pathinfo` is expected to return negative error codes, however, so when we get a positive error code from the server, we can negate it and return it to the user:
```suggestion
	if (rv < 0) {
		__simple_printf("dserver_rpc_get_executable_path failed internally: %d\n", rv);
		__simple_abort();
	} else if (rv > 0) {
		return -rv;
	}
```

facekapow (changes requested):
Looks good, but also requires some changes discussed in the darlingserver PR.

facekapow:
Like I said in the darlingserver PR, there are things you wouldn't know to do because I simply haven't documented them (e.g. the meaning of negative error codes from RPC calls).

trungnt2910:
Fixed.

facekapow (approved):
LGTM, just need to update the darlingserver commit to point to the merge.

PR #1153, opened 2022-04-07
Vision framework stubs

mrolappe:
This adds Vision framework stubs, generated on Catalina.


mrolappe:
@facekapow can this be merged now?

PR #1183, opened 2022-06-16
copyright 2016-2020 -> 2016-2022

bquast:


CuriousTommy:
@bquast Please don't make a separate PRs for a similar PR that you opened earlier. I going to close this one.

https://github.com/darlinghq/darling/pull/1182

CuriousTommy:
> Could you please offer some guidance on which copyright notices should be updated? I want to make it comprehensive, but also correct.

I would focus only on the code that was written from scratch. We generally want to make as few modification as possible on Apple's code. For this PR, just focus on the code that is in this repo.

PR #1200, opened 2022-07-29
Add Bom framework

mrolappe:
This adds Bom framework stubs, generated on Catalina.


CuriousTommy:
The Bom framework is private, it should go under `src/private-frameworks`.

On a side note, I'm currently making some changes to the cmake scripts so that the open source SDK can be manually regenerated. I want to finish that first before I merge this in.

mrolappe:
I see. Right, thanks for the hint, I will move it to `src/private-frameworks`.

CuriousTommy:
@mrolappe I made some changes to the `darling-stub-gen` program. When you get the chance, download the updated version.

CuriousTommy:
I also updated the instructions on the darling docs on how to generate the stub. Let me know if you find the updated instructions are confusing.

mrolappe:
@CuriousTommy Sorry for the late reply. Thanks for letting me know about the updated `darling-stub-gen`.

Issue #1206, opened 2022-08-21
Tracking WSL1 support


trungnt2910:
Opening this issue to track WSL1 support.

According to the [Darling docs](https://docs.darlinghq.org/wsl-build.html):
> WSL1 is more complicated and is not currently working.

There is no explicit statement from Darling denying support of WSL1, so this issue should be valid.

Currently, there are a few main issues:
- The `pidfd_open` syscall (since Linux 5.3) and the `mmap` `MAP_FIXED_NOREPLACE` flag (since Linux 4.17) is not available on Linux 4.4.0 (WSL1 kernel).
- https://github.com/microsoft/WSL/issues/8095
- https://github.com/microsoft/WSL/issues/8742
- https://github.com/microsoft/WSL/issues/8748

**System Information**

| Software | Version |
| --- | --- |
| Linux Kernel | 4.4.0-22000-Microsoft |
| Darling | e385bb9d9898987119c6514d01d403204e00704d |

trungnt2910:
At the time of writing, when #1209 is merged, `darling shell` can now run `launchd` and spawn essential system daemons, including `shellspawn`.

Very simple binaries, such as `uname`, can be loaded and run:
![image](https://user-images.githubusercontent.com/57174311/186563133-2e42f6d8-6f89-404e-bb44-a8ebb3fb6794.png)

Other binaries, including essential ones such as `bash` and `g++`, cannot run due to a hanging `mach_msg_overwrite` call. Any help here is appreciated.

trungnt2910:
`xtrace` works well, and with its help, more details about the `bash` hang can be obtained:
```
trung@DESKTOP-5OCA2N2:~$ sudo -E strace -u trung -v -s 4096 -ff -o /home/trung/darlingtrace/trace.txt darling /home/trung/.darling/usr/bin/xtrace bash
[sudo] password for trung:
xtrace: failed to dlsym(/usr/lib/darling/xtrace-mig/libnotify_xtrace_mig.dylib, "xtrace_mig_subsystem"): dlsym(0x7f0d3b9b7f70, xtrace_mig_subsystem): symbol not found
[45] sigprocmask(1, NULL, 0x7FFFFFDFFD30) -> 0
[45] sigaltstack(NULL, 0x7FFFFFDFFD20) -> 0
[45] open("/dev/tty", O_RDWR|O_NONBLOCK, 996530944) -> 3
[45] close(3) -> 0
[...]

PR #1209, opened 2022-08-25
feat: Some more WSL support

trungnt2910:
Part of https://github.com/darlinghq/darling/issues/1206.

- sendmsg: Fix a problem when an uninitialized Linux msg_name pointing to garbage can cause WSL1 to return EINVAL, even when msg_namelen is 0.
- listen: Register listening fds to libkqueue as a workaround for a problem with SO_ACCEPTCONN in WSL1.
- Updated darlingserver and libkqueue submodules with fixes.


facekapow:
Would you mind updating the darlingserver and libkqueue submodules to the merged commits?

PR #1229, opened 2022-10-27
bump cmake minimum required version to 3.13

t0rr3sp3dr0:
Since `darlingserver` requires CMake 3.13 (see darlinghq/darlingserver#6) and it is a dependency of this project, you must have CMake 3.13 or newer to build `darling`.


CuriousTommy:
@t0rr3sp3dr0 When you get the chance, can you update the sub-module as well?

PR #1292, opened 2023-02-05
Increase Cmake minimum version to 3.13, bump darlingserver submodule version

IsaccBarker:
Does virtually the same thing as #1229. That PR is stagnant, however, as the author has not responded by bumping a sub module version. This PR bumps the module version. Useful output from `git submodule status | grep darlingserver`:

```
 697fa6e643f0d7c44f5db2143f266345efb37136 src/external/darlingserver (heads/main)
```

While commit `697fa6e643f0d7c44f5db2143f266345efb37136` is the latest, and is after the commit where darlingserver got updated to version 3.13 (see commit `e9e0a9f9a4acdcc0150e90dbb0ed92f107904ebe`), the delta code change between the two shouldn't break anything as the only things changed were bug fixes (one for int-conversion and one to use explicit casting).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,7 +44,7 @@ set(CMAKE_OSX_DEPLOYMENT_TARGET 10.15 CACHE STRING "The version of macOS we're s
 
 project(darling)
 
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.13)
 enable_language(ASM)
 
 set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")


CuriousTommy:
There is another reason why I haven't merged https://github.com/darlinghq/darling/pull/1229. There are a lot of CMakeLists.txt files that have `cmake_minimum_required(VERSION ` in them.

If we are going to update the minimum version, I want to update all of the `CMakeLists.txt` files to match the `3.13` version. However, I'm currently in the process of working on the macOS 11.5 source updates (I don't want to deal with updating a bunch of sub-modules just for a cmake version change, I rather deal with it after the source updates).

CuriousTommy:
I'm going to keep this opened as a reminder to myself to update the cmake minimum version number.

PR #1340, opened 2023-03-17
Add definitions for missing constants in CoreMIDI

dingari:


CuriousTommy:
I was thrown off that these values are null, but when printed out on my Mac, these values are indeed null...

dingari:
Yes, me too!

PR #1341, opened 2023-03-17
Add stubs/missing symbols for UnicodeConverter

dingari:


CuriousTommy:
I'm looking through some of your PRs and I noticed that some of header declarations/comments are direct copies from Apple's proprietary headers.

Please don't copy and paste from Apple's proprietary header! Unless the header file **explicitly** states that is okay to do. **This is a legal liability.**

I recommend sticking with the [Apple's Developer website](https://developer.apple.com/documentation/coreservices/1433583-conversion_flags) when you are trying to implement the declarations.

dingari:
I'm truly sorry for being so sloppy! I'll look them over and make sure they're OK.

Issue #1074, opened 2021-12-14
Allow building without `python2`


ShamrockLee:
Considering that the support of Python 2 by the Python Foundation was ended at 2020-01-01, and some platforms like Nixpkgs are planning to drop the Python 2 support ([NixOS/nixpkgs #148779](https://github.com/NixOS/nixpkgs/issues/148779)). It would be a good idea to migrate to Python 3, or use PyPy2 instead.

https://www.python.org/doc/sunset-python-2/

facekapow:
Well, we can't just drop Python 2 support. I agree that Python 2 should be long gone by now, but as long as macOS supports it, Darling has to support it as well. Actually, even if macOS were to drop Python 2 support, we would probably want to keep supporting it for backwards compatibility.

As for distros removing Python 2, Darling really only needs the host system's Python 2 at build time in order to precompile the Python 2 sources included in Darling. At runtime, the Python 2 used is the one built with Darling, not the host's Python 2. All we really need from the host's Python 2 is `pyc` compilation. I prefer if they're generated from source at build time; however, if there's no other choice, we could precompile the `pyc` files ourselves and include them in a Git repo (like we do for the Swift libraries).

ShamrockLee:
As far as I know, PyPy2 should run the build scripts just fine as long as they are in pure Python (with no third-party, C-binded modules). It would be great if there's an option to specify the build-script interpreter being used.

As for `pyc` compilation, PyPy2 inherently produces different result than CPython does. It sounds like a good idea for those `*.pyc` bytecodes to be included in the source code (along with some bootstrapping instructions).

soloturn:
macos removed python2 support in macos monterrey.

Cactus-man:
I do not think most users need `python2`, considering it not being officially supported. It might be a possibility to make it an optional feature. [Ideally not by using an optional dependency](https://michael.stapelberg.ch/posts/2019-05-23-optional-dependencies/).
Maybe changing the approach to use the host `python2` at runtime could allow usage without having the build time dependency.

LubosD:
@Cactus-man We can't use host Python, because that Python is supposed to be for macOS and not for Linux.

ShamrockLee:
@LubosD Could we make `python2` an optional build-time dependency, then?

facekapow:
Thanks to @zhaofengli in #1356, it should now be possible to build without Python 2 at all.

PR #1356, opened 2023-04-23
cmake: Make Python 2 bytecode pre-compilation optional

zhaofengli:
This affects performance while preserving compatibility.

We can't remove it altogether just yet since there are still things depending on it, like older versions of Xcode. A new prefix has `xattr` and `xcode-select --install` depending on Python 2.

Ref: #1074

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,7 +30,6 @@ endif(NOT DEFINED CMAKE_CXX_COMPILER)
 
 option(DARLING_NO_CCACHE "Disable ccache usage" OFF)
 
-find_package(Python2 REQUIRED COMPONENTS Interpreter)
 find_program(CCACHE_PROGRAM ccache)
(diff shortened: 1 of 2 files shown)


facekapow (approved):
LGTM

In my opinion, this seems like a good solution for now; you're right that it *shouldn't* cause any issues, just make it execute slower. I'll wait for input from one of the other maintainers, but I certainly prefer this over including the `pyc` files in the repos.

LubosD (approved):
LGTM

bugaevc:
How about rather than making this an off-by-default option (which nobody is going to enable), we detect whether Python 2 is present and auto-enable pyc in that case?

bugaevc:
Ah, it's on by default. Fine then, though autodetection (i.e. without `REQUIRED` and then ` if (Python2_Interpreter_FOUND)`) would still be nice.

PR #1363, opened 2023-04-28
Add AE TypeDefs for AppKit Apps

CKegel:
adds required typedefs in AppleEvents so that TextEdit will compile and run

### DIFF
--- a/src/frameworks/CoreServices/include/AE/AE.h
+++ b/src/frameworks/CoreServices/include/AE/AE.h
@@ -6,4 +6,20 @@
 OSErr AECreateDesc(DescType typeCode, const void *dataPtr, Size dataSize, AEDesc *result);
 OSErr AEDisposeDesc(AEDesc *theAEDesc);
 
+enum : AEEventID {
+kAEOpenApplication = 'oapp',
+kAEReopenApplication = 'rapp',
+kAEOpenDocuments = 'odoc',
+kAEPrintDocuments = 'pdoc',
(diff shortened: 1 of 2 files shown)


src/frameworks/CoreServices/include/AE/AE.h
```diff
@@ -6,4 +6,20 @@
...
+kAEQuitApplication = 'quit',
+kAEAnswer = 'ansr',
+kAEApplicationDied = 'obit',
+kAEShowPreferences = 'pref',
+};
+
+enum : DescType {
+    kCoreEventClass = 'aevt',
+};
+
```
CuriousTommy, line 24: These declarations should be in `AppleEvents.h` instead. That's where they live in Apple's SDK.

CuriousTommy (changes requested):

src/frameworks/CoreServices/include/AE/AE.h
CKegel: Does `AppleEvents.h` currently exist, or should I create it in `src/frameworks/include/AE/`?

CuriousTommy: You can create one at `src/frameworks/include/AE`.

CKegel: I think I've done it correctly - let me know if anything else needs changed!

CuriousTommy:
Thank you for you contribution!

PR #1406, opened 2023-07-20
Implement Buffer Support For Logging In xtrace

CuriousTommy:
# Tasks
* Rewrite code to not rely on `std::string` & `std::vector`
  * Ariel's comment on why we need to do this: https://github.com/darlinghq/darling/commit/204eabb10990ca6466f7a45a116e68c732d94ea7#r122230496
* Figure out why Multi-threaded test case fails

# Test Cases

<details open>
<summary>Single Thread</summary>

```c
// single_thread_test.c
#include <stdio.h>

int main() {
    printf("Hello world ");
    #if __i386__
        printf("from i386!\n");
    #elif __x86_64__
        printf("from x86_64!\n");
    #endif
}
```
</details>

<details open>
<summary>Multi-threaded</summary>

```c
// multithreaded_thread_test.c
#include <stdio.h>
#include <pthread.h>

void* hello_thread(void* args) {
    int* value = (int*)args;
    printf("Hello from thread %d!\n", *value);
    return NULL;
}

int main() {
    printf("Hello world ");
    #if __i386__
        printf("from i386!\n");
    #elif __x86_64__
        printf("from x86_64!\n");
    #endif

    printf("Attempting to create threads\n");
    pthread_t threads[4];
    int temp[4];
    for (int i=0; i<4; i++) {
        temp[i]=i+1;
        if (pthread_create(&threads[i],NULL,hello_thread,&temp[i])) {
            printf("Unable to create thread\n");
            return -1;
        }
    }

    printf("Joining threads\n");
    for (int i=0; i<4; i++) {
        pthread_join(threads[i],NULL);
    }

    return 0;
}
```

</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ You can uninstall and list packages with the `uninstaller` command.
 
 ### Working with DMG images
 
-DMG images can be attached and deattached from inside `darling shell` with `hdiutil`. This is how you can install Xcode along with its toolchain and SDKs (note that Xcode itself doesn't run yet):
+DMG images can be attached and detached from inside `darling shell` with `hdiutil`. This is how you can install Xcode along with its toolchain and SDKs (note that Xcode itself doesn't run yet):
 
 ````
 Darling [~]$ hdiutil attach Xcode_7.2.dmg
--- a/src/libsimple/src/lock.c
+++ b/src/libsimple/src/lock.c
@@ -226,7 +226,7 @@ void libsimple_once(libsimple_once_t* _once, libsimple_once_callback callback, v
 				return;
 			}; // not reached
 			case libsimple_once_state_triggered_uncontended: {
-				// somebody is already perfoming the callback;
+				// somebody is already performing the callback;
 				// now we're waiting, so let's update the state for that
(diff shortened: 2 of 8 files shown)


CuriousTommy:
@facekapow 

> Now that we have access to basically all of the host libc (since it's loaded implicitly by mldr), we could just add elfcalls for the host's malloc and free and use those instead.

Is there an example I can look at on how I should do this? Am I suppose to use `wrap_elf` and `dlopen`?

facekapow:
@CuriousTommy A fairly simple example would be 176375e9345c4499c0970b496d956516fd7bed0b (minus the changes to `src/kernel/emulation/linux/misc/proc_info.c`, of course). No, you don't need `wrap_elf` or `dlopen`. You basically just need to add entries for `malloc` and `free` to the `struct elf_calls` structure (in `src/startup/mldr/elfcalls/elfcalls.h`) and initialize those members in `elfcalls_make` (in `src/startup/mldr/elfcalls/elfcalls.c`).

Since this is just for xtrace to use them, you don't even need to add the corresponding wrappers in libsystem_kernel; just include `<elfcalls.h>` and add an `extern struct elf_calls* _elfcalls` declaration.

PR #1445, opened 2023-10-25
RPM Improvements

bxgaillard:
Various improvements to RPM building, including:
- better compliance to RPM standards;
- support for building in a rootless container;
- use of standard POSIX shell instead of Bash;
- compatibility with recent distributions by removing build dependency to Python 2 (still installed when available);
- support for EL (RHEL, CentOS, Rocky Linux, AlmaLinux...) in addition to Fedora.

I made several small commits, feel free to squash if appropriate.


Kyr4l:
> support for EL (RHEL, CentOS, Rocky Linux, AlmaLinux...) in addition to Fedora.
does that include OpenSUS ?

PR #1504, opened 2024-04-02
Remove extra `extern` declarations from frameworks in Darling

CKegel:
Certain frameworks define constants as external in both their header and implementation file. While this is not an error, it generates a significant amount of warnings that may obscure more relevant information during the build process. This PR removes those instances of redundant `extern` keywords.


CuriousTommy (approved):
Looks good to me!

PR #1554, opened 2024-11-13
Disable dependency file generation

johnpanos:
The build fails on cmake v3.31.0, as it starts defaulting `CMAKE_${LANG}_LINKER_DEPFILE_SUPPORTED` to `TRUE` by default. Having that set to `TRUE` will make cmake add `--dependency-file` to the linker flags, however ld64 in cctools does not support this flag, causing the build to break.

This PR disables dependency file generation to match the previous behavior.


CuriousTommy:
Thank you for your contribution!

PR #1584, opened 2025-05-08
fix: stop all running darling servers on uninstall

### LINKED ISSUE #1579: tools/uninstall does not stop _all_ running darling instances.
**Expected Result**
I've got three users, and when I run `tools/uninstall` I expect it to stop the darling server across all users, much like `tools/shutdown-user.sh` does.

**Actual Result**
Darling is stopped on current user, and uninstalled (but still running for other users).

**Steps To Reproduce**
1. Standard build process
2. Standard install process
3. `darling shell` then exit in multiple users
4. Run `tools/uninstall`

(as an aside, to fix this, simply do `make install` as in the build guide, run the `shutdown-user.sh` script, then `make uninstall`)

**System Information**
What system are you using?

| Software | Version |
| --- | --- |
| Linux Kernel | 6.12.23-1-lts |
| Darling | 83c14ddd5540dd15ab3a3ca91943ead0b85415c0 |

Lazerbeak12345:
closes #1579

### DIFF
--- a/tools/shutdown-user.sh
+++ b/tools/shutdown-user.sh
@@ -1,15 +1,26 @@
+#!/bin/sh
+
+if [ "$(whoami)" != "root" ]
+then
+	echo "$0 must be run as root, invoking sudo"
+	exec sudo su -c "$0" "$@"
+fi
+
 echo "Seeing if Darling is currently running"
 
-PID=`pgrep launchd`
+# Sometimes `pgrep launchd` doesn't work when `pgrep -f launchd` does. Not sure
+# if we want to include arguments (and capture rare, but possible false
+# positives).
+PID=$(pgrep -f launchd)
 
-while [[ ! -z $PID ]]
+while [ -n "$PID" ]
 do
 	THISPID=$(echo $PID | head -n1)
 	RUNNING_USER=$(ps -o uname= -p $THISPID | head -n1)
 	echo "Darling currently running for $RUNNING_USER, shutting it down..."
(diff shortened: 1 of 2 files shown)


Lazerbeak12345:
closes #1579

CuriousTommy (changes requested):
The script is broken

```
Seeing if Darling is currently running
error: list of process IDs must follow -p

Usage:
 ps [options]

 Try 'ps --help <simple|list|output|threads|misc|all>'
  or 'ps --help <s|l|o|t|m|a>'
 for additional help text.

For more details see ps(1).
Darling currently running for , shutting it down...
sudo: unknown user darling
sudo: error initializing audit plugin sudoers_audit
```

Lazerbeak12345:
Tested much more extensively this time.

I noticed that sometimes `pgrep launchd` doesn't work when `pgrep -f launchd` does. Not sure if I want to include arguments (and capture rare, but possible false positives).

Lazerbeak12345:
The other thing: this script makes the assumption that a user can run `sudo sudo -u other_user tools/shutdown-user.sh`. Running `sudo` as the root user isn't possible by default on some systems (such as mine, which is actually using doas), but I'm pretty sure it's possible by default on fedora and ubuntu.

I may want to do a PID check first instead of relying on sudo capabilities

CuriousTommy (approved):
LGTM. Once you combine your commits into one, I'll merge your changes.

CuriousTommy:
Thank you for your contribution!

Issue #1587, opened 2025-05-11
Missing `_CreateNibReference` symbol


Lazerbeak12345:
**Expected Result**
What did you expect to happen?

The application would not error on missing this symbol

**Actual Result**
What did happen?

Inkscape 0.92.2 (an Intel Darwin X11 app) has this in its logs before a coredump:

```log
dyld: lazy symbol binding failed: Symbol not found: _CreateNibReference
  Referenced from: /Applications/Inkscape.app/Contents/MacOS/Inkscape
  Expected in: /System/Library/Frameworks/Carbon.framework/Versions/A/Carbon

dyld: Symbol not found: _CreateNibReference
  Referenced from: /Applications/Inkscape.app/Contents/MacOS/Inkscape
  Expected in: /System/Library/Frameworks/Carbon.framework/Versions/A/Carbon

abort_with_payload: reason: Symbol not found: _CreateNibReference
  Referenced from: /Applications/Inkscape.app/Contents/MacOS/Inkscape
  Expected in: /System/Library/Frameworks/Carbon.framework/Versions/A/Carbon
; code: 4
Abort trap: 6 (core dumped)
```

This is in the Inkscape source [here](https://gitlab.com/inkscape/inkscape/-/blob/0.92.x/packaging/macosx/ScriptExec/main.c#L410)
- next frame up [here](https://gitlab.com/inkscape/inkscape/-/blob/0.92.x/packaging/macosx/ScriptExec/main.c#L176)

**Steps To Reproduce**
If possible, what steps can you take to reproduce the issue?

1. Get `Inkscape-0.92.2-1-x11-10.7-x86_64.dmg` [from the official release here](https://inkscape.org/release/0.92.2/mac-os-x/). Be advised that this version is _not_ the latest version of Inkscape, and is outdated. I'm using it for testing purposes only (and because it's FOSS).
2. Install the dmg [in the usual fashion](https://docs.darlinghq.org/what-to-try.html#attach-disk-images)
3. Run this: `/Applications/Inkscape.app/Contents/MacOS/Inkscape`

**System Information**
What system are you using?

| Software | Version |
| --- | --- |
| Linux Kernel | `6.12.23-1-lts` |
| Darling | fd4ba4017c72d34244a26784f6183b79322f732f |

This is a follow-up to #1577

Lazerbeak12345:
Further research:

[...]

PR #1589, opened 2025-05-17
add missing IBCarbonRuntime.h, associated stubs and modifications to HIToolbox

Lazerbeak12345:
closes https://github.com/darlinghq/darling/issues/1587

- [x] build this
- [x] test this
- [x] squash

Uses information from https://leopard-adc.pepas.com/documentation/Carbon/Reference/Interface_Builder/Reference/reference.html and https://developer.apple.com/library/archive/releasenotes/Carbon/HIToolbox.html


src/frameworks/Carbon/HIToolbox/include/HIToolbox/IBCarbonRuntime.h
```diff
@@ -0,0 +1,37 @@
+#ifndef _Carbon_IB_Carbon_Runtime_H_
+#define _Carbon_IB_Carbon_Runtime_H_
```
Lazerbeak12345, line 2: I don't know if this is the correct name here, but I'm confident in the filename and path.

src/frameworks/Carbon/HIToolbox/include/HIToolbox/IBCarbonRuntime.h
```diff
@@ -0,0 +1,37 @@
...
+
+#include <CoreFoundation/CoreFoundation.h>
+#include <CarbonCore/MacErrors.h>
+#include <CoreFoundation/CFBase.h>
+#include <HIToolbox/Menus.h>
+#include "../../src/MacWindows.h" // TODO: put WindowRef where it actually goes
```
CuriousTommy, line 8: I'm not sure why `MacWindows.h` is not included in the header folder...

Can you create a separate commit for moving `MacWindows.h` to the header folder (and fixing the includes that reference `MacWindows.h`)?

CuriousTommy (changes requested):

CuriousTommy:
> In this PR, I've figured since the rest of the HIToolbox APIs are compiled to be available in darling, even in 64 bit intel, it should be fine to replicate that behavior for this API as well.

I think that is okay. If that becomes a problem, we can always fix the code in the future.

src/frameworks/Carbon/HIToolbox/include/HIToolbox/IBCarbonRuntime.h
Lazerbeak12345: Two problems:

- WindowRef is supposed to be in AppKit, and I'm not sure what it's doing here
- `<HIToolbox/MacWindows.h>` isn't in any list of header files I've found (I was thinking I could rename it to `<HIToolbox/_MacWindows.h>`, so it's really not much of a problem

Lazerbeak12345: found documentation for MacWindows.h. https://web.archive.org/web/20160206202121/http://mirror.informatimago.com/next/developer.apple.com/documentation/Carbon/reference/Window_Manager/wind_mgr_ref/function_group_21.html

Lazerbeak12345: I was going to just add all stubs to that header since I was there anyway, but as it turns out in Leopard, theres more than 500 symbols, so I'll not do that in this PR.

CuriousTommy: > 500 symbols, so I'll not do that in this PR

Honestly, it would be a nightmare for me to verify all 500 symbols at once... I rather do it when we find an app that needs them.

CuriousTommy (approved):
LGTM! Thanks

Issue #1596, opened 2025-05-26
[Issue Tracker] SubEthaEdit application


jlucfarias:
Based on issue #1594 of @Lazerbeak12345 and issue #1333 of @mbodomi, I will try in this issue to track all symbols and unimplemented functions to make [SubEthaEdit](https://subethaedit.net/) application work. It is not a simple app (mainly to me that never touched objective-c code before), but it will allow to some other applications to work too when all get implemented.
I will send some PRs slowly to add symbols and try to implement functions to make it work.
All symbols below was getted after every iteration of add symbol and rebuild darling to see the progression of application work.

<details><summary>Symbols and Functions</summary>
<p>

- [ ] `kSCDynamicStorePropNetPrimaryInterface`
- [ ] `kSCPropNetIPv4Addresses`
- [ ] `kSCPropNetIPv4SubnetMasks`
- [ ] `WebElementImageKey`
- [ ] `WebElementLinkTitleKey`
- [ ] `WebElementLinkTargetFrameKey`
- [ ] `WebElementLinkURLKey`
- [ ] `_OBJC_CLASS_$_NSScriptCoercionHandler`
- [ ] `_OBJC_CLASS_$_NSScriptSuiteRegistry`

</p>
</details>

**Observation:** I will edit this description to add more symbols and PRs after symbols when opened

jlucfarias:
After symbols addition the application stuck at these errors:

```sh
2025-06-25 11:02:25.165 SubEthaEdit[34:3e8] Unable to load main nib file MainMenu
2025-06-25 11:02:26.678 SubEthaEdit[34:3e8] convertFont:toHaveTrait: failed, San Francisco 2
2025-06-25 11:02:26.684 SubEthaEdit[34:3e8] convertFont:toHaveTrait: failed, San Francisco 1
2025-06-25 11:02:26.693 SubEthaEdit[34:3e8] convertFont:toHaveTrait: failed, San Francisco 1
2025-06-25 11:02:26.705 SubEthaEdit[34:3e8] Terminating app due to uncaught exception 'NSException', reason: '-[NSDocumentController isOpeningUntitledDocument]: unrecognized selector sent to instance 0x74d9e9069410'
libc++abi: terminating with uncaught exception of type NSException
Abort trap: 6 (core dumped)
```

[...]

PR #1605, opened 2025-07-16
[WIP] Add UIFoundation framework to read UIKit NibArchive files

jlucfarias:
# It is not done yet!

As of trying to make [SubEthaEdit application work](https://github.com/darlinghq/darling/issues/1596), I faced two main issues (first is focused here as it is more urgent):
1. `Unable to load main nib file X`
2. `convertFont:toHaveTrait: failed, X`

The nib file is loaded at `loadNibNamed` method from [`NSNibLoading.m`](https://github.com/darlinghq/darling-cocotron/blob/master/AppKit/nib.subproj/NSNibLoading.m) that internally call `loadNibFile` method. The latter method create an instance of NSNib and call `instantiateNibWithExternalNameTable` method that always return false for the nib file of SubEthaEdit. It is that because SubEthaEdit was compiled using Xcode 13+ that compile nib files using a new format called UIKit NibArchive that use the class UINibEncoder from UIFoundation private framework. NSNib is using right now [NSKeyedUnarchiver to read nib file](https://github.com/darlinghq/darling-cocotron/blob/master/AppKit/nib.subproj/NSNib.m#L180), but the new format should be read using UINibDecoder.

Most of implementation are derived from [this article](https://www.mothersruin.com/software/Archaeology/reverse/uinib.html) that helped me a lot to understand the new format and [this article](https://github.com/matsmattsson/nibsqueeze/blob/master/NibArchive.md) helped me on first steps.

The current state of the implementation is: it can read a nib file, get keys, values, class names and objects properly. After fix all current known issues and implement `decodeObjectForKey` method (only used in NSNib), I will try to use it on NSNib as a fallback of NSKeyedUnarchiver.

I think it can be one step to make more modern applications to work on darling.


jlucfarias:
The previous versions used only Foundation and worked on real MacOS, but not on Darling. I've had to make some refactors to be used. I did pushed only a part of the changes because I need to figure out a way to decode the first object searched in nib loading context (IB.objectdata) and to re-add NIBEncoder/Decoder to Appkit again

jlucfarias:
Closing because local changes are too complex and I need more investigations to make correct changes

PR #1631, opened 2025-11-10
Add symbols for BibDesk

jlucfarias:


src/frameworks/CoreServices/src/SearchKit/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ add_framework(SearchKit
 
 	SOURCES
 		Analysis.m
+		SKIndex.m
```
sirnacnud: Looking at the `SearchKit` headers from Apple, the library is implemented as a C library and not Objective-C.

It would probably be best if we also implement it as a C library.

CuriousTommy: > It would probably be best if we also implement it as a C library.

I don't think it really matters that much. Building C code in an Objective-C source file shouldn't affect the C code.

jlucfarias: Fixed

Issue #1489, opened 2024-02-13
Don't Use DYLD_ROOT_PATH in (darlingserver, mldr, and vchroot) 


CuriousTommy:
# Background

It turns out that `dyld` uses the `DYLD_ROOT_PATH` environment variable. 

https://github.com/darlinghq/darling-dyld/blob/63f667cf06d7ed59553adebb0c8d70a117135ac9/src/dyld2.cpp#L6598-L6612
```c
	// if this is host dyld, check to see if iOS simulator is being run
	const char* rootPath = _simple_getenv(envp, "DYLD_ROOT_PATH");
	if ( (rootPath != NULL) ) {
		// look to see if simulator has its own dyld
		char simDyldPath[PATH_MAX]; 
		strlcpy(simDyldPath, rootPath, PATH_MAX);
		strlcat(simDyldPath, "/usr/lib/dyld_sim", PATH_MAX);
		int fd = dyld3::open(simDyldPath, O_RDONLY, 0);
		if ( fd != -1 ) {
			const char* errMessage = useSimulatorDyld(fd, mainExecutableMH, simDyldPath, argc, argv, envp, apple, startGlue, &result);
			if ( errMessage != NULL )
				halt(errMessage);
			return result;
		}
	}
```

This is an issue since `darlingserver`, `mldr`, and `vchroot` also use the same environment variable.

https://github.com/darlinghq/darlingserver/blob/62a3321e98cbc9bc9d52c8f6f0d588e65c789954/src/darlingserver.cpp#L269
```c
	setenv("DYLD_ROOT_PATH", LIBEXEC_PATH, 1);
```

https://github.com/darlinghq/darling/blob/25afbc76428c39c3909e9efcf5caef1140425211/src/startup/mldr/mldr.c#L522-L529
https://github.com/darlinghq/darling/blob/25afbc76428c39c3909e9efcf5caef1140425211/src/vchroot/vchroot.c#L49

# Solution

A proper fix would be to reword the `DYLD_ROOT_PATH` environment variable name (that `darlingserver`, `mldr`, and `vchroot` uses) to something that is more specific to Darling (ex: `DARLING_DYLD_ROOT_PATH`).

CuriousTommy:
@CKegel If you (or your team) want to work on this, I don't mind assigning this ticket to you.

CKegel:
[...]

PR #1638, opened 2025-12-06
Fix usage of DYLD_ROOT_PATH

sirnacnud:
This fixes mldr to not use `DYLD_ROOT_PATH` directly and instead reads it in via `__mldr_DYLD_ROOT_PATH`.

The changes to `darlingserver` are here: https://github.com/darlinghq/darlingserver/pull/14

The solutions is described [here](https://github.com/darlinghq/darling/issues/1489#issuecomment-3595050574).

I evolved the solution a bit to make the environment variable mdlr specific with `__mldr_` prefix rather than `DARLING_`, as this is a special case of mldr needing to setup `DYLD_ROOT_PATH` for the executable, which is only needed when executing `vchroot`.

With this fix, you can now set `DYLD_ROOT_PATH` on the dalring shell and it won't break launching further executables. This support would be needed to run an iOS simulator, which needs `DYLD_ROOT_PATH` set to the iOS system libraries.


CuriousTommy (approved):
LGTM!

PR #1710, opened 2026-03-12
Add more Bibdesk symbol

jlucfarias:


CuriousTommy (approved):
LGTM, but with one side-note...

One interesting thing is that `DOMElement` actually lives in the `WebKitLegacy.framework`, which `WebKit.framework` re-exports from.

I don't think we'll need to worry about this for now. In the long run, we are going to move away from this stub implementation and use [Apple's open source release](https://github.com/apple-oss-distributions/WebKit/tree/main).

PR #1717, opened 2026-03-14
Add PubSub Framework

sirnacnud:
`PubSub` is required for Xcode 3.0 and 5.0. `PubSub` was removed in Catalina, so I used Mojave to run the stub generator.

I also added symbol for `PSFeedRefreshingNotification`.

It was not that straightforward to compile a program using `PubSub` on Mojave, as `PubSub.framework` isn't part of the MacOSX10.14 SDK. I had to extern the symbol since there are no headers and setup a symlink to the `PubSub.framework` on the system.

Source (symbol-print.m):
```
#import <Foundation/Foundation.h>

extern NSString* const PSFeedRefreshingNotification;

int main() {
    NSLog(@"%@", PSFeedRefreshingNotification);
    return 0;
}
```

In `/Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk/System/Library/Frameworks`
```
sudo ln -s /System/Library/Frameworks/PubSub.framework PubSub.framework
```
Compile with clang:
```
clang -framework PubSub -framework Foundation symbol-print.m
```


src/frameworks/PubSub/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
+project(PubSub)
+
+set(DYLIB_COMPAT_VERSION "1.0.0")
+set(DYLIB_CURRENT_VERSION "1.0.0")
+set(FRAMEWORK_VERSION "A")
+
+remove_sdk_framework(PubSub
+)
```
CuriousTommy:
```suggestion
remove_sdk_framework(PubSub
    PRIVATE
)
```

src/frameworks/PubSub/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
...
+
+generate_sdk_framework(PubSub
+    VERSION ${FRAMEWORK_VERSION}
+    HEADER "include/PubSub"
+)
```
CuriousTommy:
```suggestion
generate_sdk_framework(PubSub
    PRIVATE
    VERSION ${FRAMEWORK_VERSION}
    HEADER "include/PubSub"
)
```

CuriousTommy (changes requested):
Since PubSub is a private framework, it should live in `src/private-frameworks` not `src/frameworks`.

Also header symlinks should live in `PrivateFramework`, not `Framework`

sirnacnud:
> Since PubSub is a private framework, it should live in `src/private-frameworks` not `src/frameworks`.
> 
> Also header symlinks should live in `PrivateFramework`, not `Framework`

PubSub is a public osx framework, it just was removed in Catalina. Here is the retried documentation for it: https://developer.apple.com/library/archive/documentation/InternetWeb/Conceptual/PubSub/overview/overview.html

CuriousTommy:
> PubSub is a public osx framework, it just was removed in Catalina.

I did a little more research, and the `PubSub.framework` is actually a symlink to the private framework.

```
DarlingHQ-macOS-1014:~ darling$ ls -l /System/Library/Frameworks/ | grep PubSub
lrwxr-xr-x   1 root  wheel   37 Feb 28  2025 PubSub.framework -> ../PrivateFrameworks/PubSub.framework
```

> PubSub is required for Xcode 3.0 and 5.0. 

I wonder what the what the output would be if you run `otool -L` on those binaries. Does it use the `Frameworks/PubSub.framework` path or the `PrivateFrameworks/PubSub.framework`  path.

sirnacnud:
>I wonder what the what the output would be if you run otool -L on those binaries. Does it use the Frameworks/PubSub.framework path or the PrivateFrameworks/PubSub.framework path.

It does look like `PubSub.framework` is expected in `Frameworks` rather than in `PrivateFrameworks`. I wonder if moving it to private was part of the deprecation process.

For both Xcode 3.0 and Xcode 5.0, `PubSub.framework` is needed by some `DocSetAccess` framework, with Xcode 5.0 having some seemingly newer version of the framework (`DADocSetAccess` instead of `DocSetAccess`)

### Xcode 5.0:

Darling without `PubSub.framework`:
```
$ /Applications/Xcode-5.0.app/Contents/MacOS/Xcode 
dyld: Library not loaded: /System/Library/Frameworks/PubSub.framework/Versions/A/PubSub
  Referenced from: /Applications/Xcode-5.0.app/Contents/SharedFrameworks/DADocSetAccess.framework/Versions/A/DADocSetAccess
  Reason: image not found
abort_with_payload: reason: dyld: No shared cache present
Library not loaded: /System/Library/Frameworks/PubSub.framework/Versions/A/PubSub
  Referenced from: /Applications/Xcode-5.0.app/Contents/SharedFrameworks/DADocSetAccess.framework/Versions/A/DADocSetAccess
  Reason: image not found; code: 1
Abort trap: 6 (core dumped)
```
Output of otool on OSX:
```
$ otool -L DADocSetAccess 
DADocSetAccess:
	@rpath/DADocSetAccess.framework/Versions/A/DADocSetAccess (compatibility version 1.0.0, current version 3355.0.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.0.0)
	/System/Library/Frameworks/CoreServices.framework/Versions/A/CoreServices (compatibility version 1.0.0, current version 59.0.0)
	/System/Library/Frameworks/Foundation.framework/Versions/C/Foundation (compatibility version 300.0.0, current version 1052.0.0)
	/System/Library/Frameworks/CoreData.framework/Versions/A/CoreData (compatibility version 1.0.0, current version 478.2.0)
	/System/Library/Frameworks/PubSub.framework/Versions/A/PubSub (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/Security.framework/Versions/A/Security (compatibility version 1.0.0, current version 55433.0.0)
	/System/Library/Frameworks/Quartz.framework/Versions/A/Frameworks/PDFKit.framework/Versions/A/PDFKit (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libobjc.A.dylib (compatibility version 1.0.0, current version 228.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1197.1.1)
```
### Xcode 3.0

Darling without `PubSub.framework`:
```
$ /Applications/Xcode-3.0.app/Contents/MacOS/Xcode
dyld: Library not loaded: /System/Library/Frameworks/PubSub.framework/Versions/A/PubSub
  Referenced from: /Library/PrivateFrameworks/DocSetAccess.framework/Versions/A/DocSetAccess
  Reason: image not found
abort_with_payload: reason: dyld: No shared cache present
Library not loaded: /System/Library/Frameworks/PubSub.framework/Versions/A/PubSub
  Referenced from: /Library/PrivateFrameworks/DocSetAccess.framework/Versions/A/DocSetAccess
  Reason: image not found; code: 1
Abort trap: 6 (core dumped)
```

Output of otool on OSX:
```
$ otool -L DocSetAccess 
DocSetAccess:
	@rpath/DocSetAccess.framework/Versions/A/DocSetAccess (compatibility version 1.0.0, current version 47.0.0)
	/System/Library/Frameworks/CoreServices.framework/Versions/A/CoreServices (compatibility version 1.0.0, current version 32.0.0)
	/System/Library/Frameworks/Foundation.framework/Versions/C/Foundation (compatibility version 300.0.0, current version 673.0.0)
	/System/Library/Frameworks/CoreData.framework/Versions/A/CoreData (compatibility version 1.0.0, current version 185.0.0)
	/System/Library/Frameworks/PubSub.framework/Versions/A/PubSub (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/Security.framework/Versions/A/Security (compatibility version 1.0.0, current version 31122.0.0)
	/System/Library/Frameworks/Quartz.framework/Versions/A/Frameworks/PDFKit.framework/Versions/A/PDFKit (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libgcc_s.1.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 111.0.0)
	/usr/lib/libobjc.A.dylib (compatibility version 1.0.0, current version 227.0.0)
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 473.0.0)
```

CuriousTommy:
> It does look like PubSub.framework is expected in Frameworks rather than in PrivateFrameworks. I wonder if moving it to private was part of the deprecation process.

Hmm... since macOS Mojave is the last version to include the `PubSub.framework`, let's have our implementation match how it is handled on macOS Mojave (with some minor differences):
* The source code should be moved to `src/private-frameworks`, since Mojave has the actual implementation live in the `PrivateFrameworks` folder.
* The framework should be installed in the `PrivateFrameworks` folder (I think adding the `PRIVATE` flag in `add_framework` should do that).
* It would make the most sense to have the header files live alongside the source code (in `src/private-frameworks`)
  * However, we will have the SDK treat the header files as public (we should add a note about why we are doing this in the `CMakeLists.txt` build script).
* In `src/frameworks/PubSub`, we will create a `CMakeLists.txt` that creates a relative symlink to `PrivateFrameworks/PubSub.framework` (just like how Apple does it)

sirnacnud:
@CuriousTommy I have moved PubSub to a private framework with a symlink being created for the public version. 

Only thing I did differently than what you mentioned above, was I left the SDK headers for PubSub private. They are actually not included in the Mojave 10.14 SDK, which I did mention in the PR description. I can change them to be public if you think that would be better.

CuriousTommy:
> Only thing I did differently than what you mentioned above, was I left the SDK headers for PubSub private. They are actually not included in the Mojave 10.14 SDK, which I did mention in the PR description. I can change them to be public if you think that would be better.

The reason why I wanted PubSub to be in the public headers is that there is probably an earlier version of XCode that included the header files for PubSub.

But if that is not the case, then I'm fine with PubSub being in the `private-frameworks`

sirnacnud:
> > Only thing I did differently than what you mentioned above, was I left the SDK headers for PubSub private. They are actually not included in the Mojave 10.14 SDK, which I did mention in the PR description. I can change them to be public if you think that would be better.
> 
> The reason why I wanted PubSub to be in the public headers is that there is probably an earlier version of XCode that included the header files for PubSub.
> 
> But if that is not the case, then I'm fine with PubSub being in the `private-frameworks`

Thanks for the explanation, the headers are indeed included with Xcode 3.0 and 5.0. I have changed the SDK headers to be public now.

CuriousTommy (approved):
Looks good to me!

PR #1736, opened 2026-03-17
Add clone script

amgdev9:
I coded a script (with help of AI) to help clone the project submodules depending on the component we want to compile so we reduce the time to clone, hope it helps! Tested that `core` and `system` components compile fine when cloning submodules with this script

### DIFF
--- a/tools/fetch_submodules.sh
+++ b/tools/fetch_submodules.sh
@@ -1,0 +1,387 @@
+#!/bin/bash
+
+set -e
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+PROJECT_ROOT="$(dirname "$SCRIPT_DIR")"
+EXTERNAL_DIR="$PROJECT_ROOT/src/external"
+
+if [ -z "$1" ]; then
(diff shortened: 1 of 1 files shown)


darlingtester:
> I coded a script (with help of AI) to help clone the project submodules depending on the component we want to compile so we reduce the time to clone, hope it helps! Tested that `core` and `system` components compile fine when cloning submodules with this script

using ai in code is a bad idea because it can hallucinate and introduce bugs this code could have bugs and other bad stuff also there is a chance ai could embed malicious code from malicious sources that the ai reads

an example of bugs caused by ai are the recent windows 11 updates some of them were coded by ai and used ai in some form and had bad results like erasing the drive

CuriousTommy:
Thank you for taking your time to work on this PR. Unfortunately... I think I have to reject this PR.

> Tested that `core` and `system` components compile fine when cloning submodules with this script

That may be case now, but I don't want to take on the responsibility for making sure that `core` and `system` will still build fine in the future.

I don't want to make any promises that Darling will build fine on a partial submodule clone. If anyone want to make a bug report about Darling's build system, a full clone must be a hard requirement.

PR #1753, opened 2026-05-20
ARM64 (aarch64) runtime, ABI, and build-system support

kkHAIKE:
Bring Darling to a working state on Linux/ARM64 hosts. Real macOS Mach-O binaries (including arm64e PAC-signed ones from macOS-26 CLT) load and run on this branch.

## Top-level

* `CMakeLists.txt`: detect host arch via `uname -m`, add `TARGET_ARM64` option (defaulted on for aarch64 hosts), switch the i686-fatal error message to permit aarch64.
* `basic-headers/arm/arch.h`, `cmake/architecture.cmake`, `cmake/compiler_include.cmake`: thin arch helpers so the build system actually selects the arm64 triplet.
* `Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/{architecture/arm,arm,libkern/arm,mach/arm,mach/arm64}/*`: add arm/arm64 header symlinks (mostly thin pointers to the canonical Darwin layout).
* `src/libm/`: ARM64 math runtime (`Source/ARM/arm64_builtins.c` + fenv/fpclassify/required_arithmetic, `Source/complex.h` gains an `__arm64__`/`__aarch64__` arch branch).
* `src/startup/mldr/loader.c`: force the initial Mach-O mmap into the low-VA window (<2^47) — Linux ARM64 kernels happily return 48-bit VAs, which collide with ObjC's 47-bit `FAST_DATA_MASK` and crash `class_initialize`. Retry with `MAP_FIXED` if the kernel returns a high VA, sequential slots for repeated loads.
* `src/startup/mldr/{commpage.c,elfcalls/threads.c,include/arm/*,include/libkern/arm/*,include/mach/arm/*,mldr.c}`: per-arch headers + branch-outs so mldr's translation layer compiles + bootstraps arm64.
* `src/frameworks/CoreServices/`: add `reexport_arm64.exp` so CoreServices re-exports the right symbols on arm64.
* `src/launchd/src/core.c`, `src/libelfloader/wrapgen/wrapgen.cpp`: minor arm64-conditional adjustments.

## Submodule changes (linked PRs)

Each is independent and self-contained; this umbrella PR bumps the pointers.

### Per-package session fixes
[...]

PR #1754, opened 2026-05-20
Upgrade bundled libc++ to LLVM 15.0.7

kkHAIKE:
Catch the bundled `libc++` / `libc++abi` up to LLVM 15.0.7 so modern CLT/Xcode binaries (clang 17, swift 6.x, etc.) can bind cleanly against `std::__1::*` symbols (e.g. `__libcpp_verbose_abort`) that the previous LLVM-13.0.1-based build did not export.

**This benefits both arm64 and x86_64 — the libc++ ABI catch-up applies regardless of host architecture, and the branch can be initially tested on amd64.**

## Top-level

`CMakeLists.txt`: when `CMAKE_C_COMPILER_VERSION >= 16` (Clang), append `-Wno-error=implicit-function-declaration -Wno-error=implicit-int -Wno-error=int-conversion -Wno-error=incompatible-function-pointer-types` to `IGNORED_WARNINGS`. clang 16 promoted these from warnings to errors by default; lots of Darling's vendored C source predates C99 strictness, so we demote them rather than audit every legacy file. Required because libcxx-15 needs clang 14+ to build (uses `__attribute__((using_if_exists))` in many headers) and the rest of the tree compiles fine with clang 16+ once these specific errors are suppressed.

## Submodule changes (linked PRs)

* darlinghq/darling-libcxx#4 — source bump to `llvmorg-15.0.7` (preserving Darling's `CMakeLists.txt` and `__config_site` with `_LIBCPP_ABI_NAMESPACE=__1`).
* darlinghq/darling-libcxxabi#3 — stays at `llvmorg-13.0.1` source with compat defines for libcxx-15 (the ABI between the two is Itanium-stable; avoids a libcxxabi-15 port).
* darlinghq/darling-dyld#6 — `VECTOR_NEVER_DESTRUCTED*` macros expanded to nothing (libcxx 15 refactored away `std::__vector_base` which Apple's destructor-elision trick targeted).

## Verified

* Existing tests (`hello world`, `bash`, `uname -a`) still work — non-destructive upgrade.
* Apple CLT 26.2 `clang-format --version` (`Apple clang-format version 17.0.0 (clang-1700.6.4.2)`, `built for Mac OS X 14.0`, `MacOSX26.2.sdk`) runs to completion on Darling arm64 — first modern Apple Mach-O CLT binary observed working end-to-end.
* On amd64 the changes are minimal in scope — only `dyld/src/ImageLoader.h` macro-noop, plus the libc++/libc++abi source overlays, plus the top-level warning relax. No arm64-specific code is touched.

## Toolchain requirement

This branch needs **clang 14+** at minimum (libcxx-15 sources use `__attribute__((using_if_exists))`). Tested against clang-19 from Debian backports.

This PR pairs naturally with darlinghq/darling#1753 (arm64 support). They are independent and either can land first.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,6 +112,14 @@ if(${CMAKE_C_COMPILER_ID} STREQUAL "Clang" AND ${CMAKE_C_COMPILER_VERSION} VERSI
 	set(ASM_IGNORED_WARNINGS "${ASM_IGNORED_WARNINGS} -Wno-undef-prefix")
 endif()
 
+# DARLING: clang 16+ promoted several long-standing warnings to errors by default
+# (implicit-function-declaration, implicit-int, etc.). Lots of Darling's vendored
+# C code predates C99 strictness; reverting these to warnings keeps the build
+# moving without auditing every legacy source file.
+if(${CMAKE_C_COMPILER_ID} STREQUAL "Clang" AND ${CMAKE_C_COMPILER_VERSION} VERSION_GREATER_EQUAL "16")
+	set(IGNORED_WARNINGS "${IGNORED_WARNINGS} -Wno-error=implicit-function-declaration -Wno-error=implicit-int -Wno-error=int-conversion -Wno-error=incompatible-function-pointer-types")
+endif()
+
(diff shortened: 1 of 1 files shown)


jrmuizel:
@CuriousTommy Is there anything blocking this from landing?

PR #1760, opened 2026-06-17
Add symbols for Google's Backup and Sync

jlucfarias:
Add symbols mentioned at https://github.com/darlinghq/darling/issues/342

PR #1779, opened 2026-08-31
Add source file and fix header build

jlucfarias:
The source file was missing from the build in #1760, so it wasn't compiled and an error in the header went unnoticed.

This PR adds the missing file to the build and fixes the header error.


CuriousTommy (approved):
Good catch!

Issue #1594, opened 2025-05-22
unimplemented stubs in Inkscape's Platypus wrapper


Lazerbeak12345:
**Expected Result**
What did you expect to happen?

The wrapper should be able to launch Inkscape (and then be able to forward the events)

**Actual Result**
What did happen?

The app exits gracefully, and these stubs are logged:

```log
STUB: InitCursor called
STUB: AEInstallEventHandler called
STUB: AEInstallEventHandler called
STUB: AEInstallEventHandler called
STUB: AEInstallEventHandler called
STUB: GetApplicationEventTarget called
STUB: InstallEventHandler called
STUB: CreateNibReference
STUB: SetMenuBarFromNib
STUB: DisposeNibReference
STUB: AECreateDesc called
STUB: OpenDefaultComponent called
STUB: RunApplicationEventLoop called
```

- [ ] `InitCursor`
- [ ] `AEInstallEventHandler`
- [ ] `GetApplicationEventTarget`
- [ ] `InstallEventHandler`
- [ ] `CreateNibReference`
- [ ] `SetMenuBarFromNib`
- [ ] `DisposeNibReference`
- [ ] `AECreateDesc`
- [ ] `OpenDefaultComponent`
- [ ] `RunApplicationEventLoop` 

**Steps To Reproduce**
If possible, what steps can you take to reproduce the issue?

1. Get `Inkscape-0.92.2-1-x11-10.7-x86_64.dmg` [from the official release here](https://inkscape.org/release/0.92.2/mac-os-x/). Be advised that this version is _not_ the latest version of Inkscape, and is outdated. I'm using it for testing purposes only (and because it's FOSS).
2. Install the dmg [in the usual fashion](https://docs.darlinghq.org/what-to-try.html#attach-disk-images)
3. Run this: `STUB_VERBOSE=1 /Applications/Inkscape.app/Contents/MacOS/Inkscape`

**System Information**
What system are you using?

| Software | Version |
| --- | --- |
| Linux Kernel | `6.12.27-1-lts` |
| Darling | b7a59db91eac44f2f2b1cb4c324c28189d10dbdc |

This is a follow-up to #1590

Lazerbeak12345:
I forgot to link the inkscape source that is calling these stubs: https://gitlab.com/inkscape/inkscape/-/blob/0.92.x/packaging/macosx/ScriptExec/main.c

Lazerbeak12345:
[...]

Issue #937, opened 2021-02-25
Cocotron backend rework


LubosD:
This issue documents how Cocotron's integration to lower API layers should work.

## Backends for X11 and Wayland

* AppKit should do CGS* API calls to interface with the window server. This means that backends from `cocotron/AppKit/*.backend` must go. There's a nice reversed collection of CGS API declarations [here](https://github.com/NUIKit/CGSInternal).
* Implement new backends in `cocotron/CoreGraphics/*.backend` that implement the new CGS interfaces.
  * The X11 backend shall use the XInput2 API to receive events.

## Event Processing
* All events must exist as [CGSEventRecords](https://github.com/darlinghq/darling-cocotron/blob/master/CoreGraphics/include/CoreGraphics/CoreGraphicsPrivate.h#L98) prior to becoming CGEvents (`CGEventCreateWithEventRecord()`) and later NSEvents (for AppKit apps).
  * All CGS events run [via an event port](https://github.com/NUIKit/GraphicsServices/blob/master/GraphicsServices/GSApp.c#L92) (`CGSGetEventPort()`)
  * Call stack of event reception from the window server:
```
              CoreGraphics`CGEventCreateFromDataAndSource+0xbce
              CoreGraphics`CGSDecodeEventRecord+0x6a
              CoreGraphics`CGSDispatchDatagramsFromStream+0x28f
              CoreGraphics`snarfEvents+0x12a
              CoreGraphics`CGSGetNextEventRecordInternal+0x9f
              CoreGraphics`CGEventCreateNextEvent+0x2c
              HIToolbox`PullEventsFromWindowServerOnConnection(unsigned int, unsigned char)+0x58
              CoreFoundation`__CFMachPortPerform+0x75
              CoreFoundation`CFRunLoopRunSpecific+0xf51
              CoreFoundation`CFRunLoopRunInMode+0x58
              HIToolbox`RunCurrentEventLoopInMode+0x11b
              HIToolbox`ReceiveNextEventCommon+0x176
              HIToolbox`BlockUntilNextEventMatchingListInMode+0x6a
              AppKit`_DPSNextEvent+0x291
              AppKit`-[NSApplication nextEventMatchingMask:untilDate:inMode:dequeue:]+0x80
              AppKit`-[NSApplication run]+0x31b
```

Another call stack:

```
1 com.apple.framework.IOKit 0x98832c8e IOHIDEventCreateWithBytes + 136
[...]

PR #1780, opened 2026-09-02
WIP feat: implement Carbon EventLoop

Lazerbeak12345:
Lots of bugs. Only partially implemented.

# TODO:

1. [ ] What exact header should the function be in?
2. [ ] Move the the function into both that .h and it's corresponding .c file
3. [ ] Import the part of CoreFoundation we need (Or just all of it, idk)
4. [ ] Call CFRunLoopRun
5. [ ] What happens when I launch Inkscape at this point? Is this the correct behavior?
6. [ ] Draft a justification of the above, using CFRunLoop and RunApplicationEventLoop documentation.
   Provide a sense of direction for how InstallEventHandler & co. will work.

# What is RunApplicationEventLoop?

- It:
  1. [ ] "Installs the standard application event handler"
  2. [ ] Suspends the app till the next event
  3. [ ] Enqueues events
  4. [ ] Dispatches enqueued events

# What is CFRunLoop?

- It's an object
- Calling CFRunLoopRun does:
  - Gets events from either _Input Sources_ or _Timer Sources_
  - Different modes get events from different places, this function starts the default mode
  - Input sources can be
    - Port based (Mach ports)
	  - 
	- Custom input sources from another thread
  - Order of events
    1. Notify entry
	2. Notify of upcoming ready timers
	3. Notify that non-port input sources will fire
	4. Fire non-port inputs
	5. If port based source is ready and waiting, process & go to 9
	6. Notify that thread will sleep
	7. Sleep till:
	  - Event arrives for port
	  - timer fires
	  - run loop timeout expires
	  - run loop explicity woken
	8. Notify observers thread has woken
	9. process pending event
	10. Notify that run loop has exited

Makes me think it may need to be in a loop

On multithreading:

> The RunApplicationEventLoop function works only on the main event loop; if your application creates
preemptive threads, each of them will have its own event loop and queue, and you must retrieve and dispatch
these events manually. For more information, see “Processing events manually” (page 41).

- [ ] play nice with the multithreading model

> Each event loop is represented by an event loop reference, an opaque data object of type EventLoopRef.
A thread can obtain a reference to its own event loop or to the program’s main event loop by calling the
functions GetCurrentEventLoop or GetMainEventLoop, respectively.

- [ ] this pr might actually work out better in the long run if I implement these as well.


Lazerbeak12345:
Makes progress toward https://github.com/darlinghq/darling/issues/1594

LubosD:
Not sure what your plan is, but when I was looking into this a few years back, I collected various backtraces that showed how Carbon and Cocoa were layered/intertwined so that the reimpl follows the same architecture.

You can find it here: https://github.com/darlinghq/darling/issues/937

Lazerbeak12345:
Thanks! I had forgotten that you had deduced the pipeline and gathered the coredumps.

I'll try to implement this layer to reflect that. I definitely will have to call stubs in the layer below what I'm working on. I must find some limit to what I implement.

I've been pulling from a variety of documentation sources. I'll link them in the OP.

Lazerbeak12345:
I've given this some thought and I've determined it'd be best if I first implemented a series of pull requests to the testsuite against many of these functions, and for any that don't exist yet (IE that I had to add to the headers) I should split into their own PRs.

This is ballooning, as I feared it would.

The least of the reasons being that Lubos is _completely correct_ that we should not just be implementing the MainEventLoop as a thin wrapper around CFRunLoop, but instead should be following Apple's pretty clear documentation on how the EventLoop works, and implementing _that_ atop of CoreFoundation's Run Loop.

My current approach would fix the problem for the Platypus wrapper, but would not help anyone using a custom event loop (which is an encouraged thing in the before-mentioned documentation from apple).

I'm not really surprised it's gone this way: I mostly put this PR here so I'd have an easier way to communicate what changes I'm working on.

PR #1787, opened 2026-09-18
fix(applications): draw grid icons upright in flipped grid view + fix NSTask stdio poisoning

cristim:
This PR includes two fixes:

1. **Icon flip fix** (from earlier): The Applications viewer's grid view was drawing icons upside-down when flipped. Fixed by correcting the coordinate transformation in main.m.

2. **NSTask stdio poisoning fix** (new): NSTask was unconditionally closing the caller's stdio handles after posix_spawn, which poisoned the process-global stdout/stderr singletons ([NSFileHandle fileHandleWithStandardOutput] / [NSFileHandle fileHandleWithStandardError]). This caused subsequent launches to fail with NSFileHandleOperationException: NSStdIOFileHandle fileDescriptor no such file or directory.

The fix:
- In NSTask.m: Skip close+dup2 file actions when the source fd equals the target fd (0, 1, or 2), avoiding EINVAL from dup2 on a just-closed fd.
- Remove the unconditional closeFile() calls after posix_spawn - NSTask should not take ownership of caller-supplied handles.

Also includes a foundation submodule update with the NSTask fix.


cristim:
Closing, this should have been created against the VibeDarling branch

